PR #5, opened 2023-07-15
Use a temp fork for tests to iterate quickly

MridulS:


MridulS:
We should probably use your fork @20kavishs so you can quickly add the dispatch decorator to the functions you require and push those changes on a dedicated branch of your fork.

You can look at https://github.com/mriduls/networkx/tree/nxp-fork as an example of how this would look like. The commit https://github.com/MridulS/networkx/commit/6d56b0a85a0457f58f6ee431cc8b4033e18a42e4 adds the required decorators for the changes in https://github.com/20kavishs/nx_parallel-1/pull/1

MridulS:
Wait on this, we may not require to do this.

PR #9, opened 2023-09-13
Reconfigure so ParallelGraph stores original nx graph

dschult:
configure ParallelGraph so that it just stores the origin graph as an attribute.

Also need `is_directed` and `is_multigraph` to call the stored graphs methods.
clean up the linting and testing workflows, pre-commits, pyproject.toml


dschult:
The test failures here come about from having a different number of `chunks` than local tests. I'm working on a system for testing which would allow setting of the number of chunks. The obvious problem of 1 chunk vs >1 chunk is most important to test, but other values might be helpful too.

PR #19, opened 2023-10-13
Add release process documentation

jarrodmillman:


RELEASE.md
```diff
@@ -0,0 +1,59 @@
...
+
+- Review the github release page:
+
+      https://github.com/networkx/nx-parallel/tags
+
+- Update `__version__` in `nx_parallel/__init__.py`.
+
+- Commit changes:
+
+      git add pyproject.toml
```
dschult, line 57: I guess this line should add `nx_parallel/__init__.py` instead of `pyproject.toml` since the previous update step is changed.

PR #25, opened 2023-12-20
replaced __networkx_plugin__ with __networkx_backend__

Schefflera-Arboricola:
ref. https://github.com/networkx/networkx/pull/7157


MridulS (approved):
Thanks!

PR #26, opened 2023-12-29
replaced networkx.plugins with networkx.backends and updated readme

Schefflera-Arboricola:
ref https://github.com/networkx/networkx/pull/7192

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,28 @@
 ## nx-parallel
 
-nx-parallel is a NetworkX backend plugin that uses joblib and multiprocessing for parallelization. This project aims to provide parallelized implementations of various NetworkX functions to improve performance.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,19 +1,28 @@
...
 
 ```
-├── centrality
-│   ├── betweenness_centrality
-│   ├── closeness_vitality
-├── tournament
-│   ├── is_reachable
-├── efficiency_measures
-│   ├── local_efficiency
+├── algorithms
```
dschult, line 10: I’m not sure this tree of directories and modules and functions is all the helpful — and it certainly is not what the heading says is “functions”.  It might be helpful to developers who want to know which functions are where, but that can be found using `git grep`.  What do you think about just having a list of the functions nx-parallel provides?

MridulS (approved):

dschult (approved):

PR #28, opened 2024-01-03
MAINT : added lint workflow test

Schefflera-Arboricola:
added `lint.yml`


dschult (approved):
This looks like it works. Did the linting tell us to change `type` to `isinstance` and to shorten the line in the doc_string?  Or are those changes we need for other reasons?

I approve this PR no matter what the answers to those questions are. :)

Schefflera-Arboricola:
>  Did the linting tell us to change type to isinstance and to shorten the line in the doc_string?

yes

Also, should I add some more tests to this pr?

Thanks!

dschult:
If the additional tests are linting related add them here -- otherwise it also makes sense to add them separately. That way it is easier to review the PRs.

Schefflera-Arboricola:
I think we can add more in future.

PR #7, opened 2023-07-16
[WIP]: Refactor-- consolidate and simplify

dPys:
This PR includes the following changes:
* Switched to absolute imports to follow best practices recommended by [PEP8](https://www.python.org/dev/peps/pep-0008/).
* Adds, consolidates, and streamlines critical functionality proposed in PR #2 (the remaining tests and algorithms for that PR will still need to be reviewed and merged)
* Adds the `Backends` class to a new module `external` for handling different parallelization backends (    "multiprocessing", "dask", "ray", "loky", "threading", and "ipyparallel") through the joblib API.
* Adds a new `optional_package` function to the `nx_parallel/misc.py` file to accommodate for the possibility of multiple backend dependencies.
* Adds a `partition` module, comprising general `NxMap` and `NxReduce` classes that can be used to more easily and consistently chunk, map, and reduce parallelizable components of most nx algorithms (as of now, "nodes", "edges", "isolates", and "neighborhoods" are supported).
* Leverage similar structure for the `algorithms` module as proposed by @20kavishs, but with the help for the `partition` classes, this greatly simplifies the implementation for each parallel algorithm variant.
* Updated the `dependencies` section in the `pyproject.toml` file to impose some minimal version requirement for NetworkX and joblib.
* Once there's consensus on general restructuring, will gladly add unit tests, and address some of the gitworkflow failures

PR #29, opened 2024-01-06
ENH : Added `get_chunks` to `betweenness_centrality` and `create_iterables` in `utils`

Schefflera-Arboricola:

1. Added `get_chunks` kwarg to `betweenness_centrality` : 
    - description: A user-defined function to create `node_chunks`(instead of how we currently do it i.e. by creating partitions on the `G.nodes`; if `get_chunks` is `"chunks"`(default) then this is executed). To understand the motivation for this see the additional pytest added
    - also added additional tests(specific to nx-parallel) to test additional kwargs
2. added `create_iterables`(ref. Pr #7)


dschult (commented):
I like this approach. It is general (and works for nodes, edges and isolated_nodes) and allows the user to override the choice of node chunks. I'm not sure how much difference it makes in betweenness centrality to be able to split up the "hard" nodes, but in general it will make a difference if you can split them up. And betweenness centrality could have some nodes take a lot longer than others. So this is a good one to start with.

I don't know of a way to test that this is working. Maybe you should have a test that computes it using the built-in default chunking and also computes it with a user-defined method. They could be "smoke-tests" to see if using that option breaks it so badly that smoke starts coming out. If smoke doesn't come out, it might still be wrong. But you have at least tested that there are no fundamental problems.  Have you thought about tests for this?

nx_parallel/algorithms/centrality/tests/test_betweenness_centrality.py
```diff
@@ -0,0 +1,33 @@
...
+        chunks = [[j[0] for j in i] for i in chunks_with_avg]
+
+        return chunks
+
+    G = nx.fast_gnp_random_graph(50, 0.5, seed=5, directed=False)
+    ebc = nx.edge_betweenness_centrality(G)
+    par_bc = nxp.betweenness_centrality(G, get_chunks=get_chunk)  # smoke test
+    bc = nx.betweenness_centrality(G)
+    for i in range(20):
+        assert round(par_bc[i], 15) == round(bc[i], 15)
```
Schefflera-Arboricola, line 33: The test was failing because of the last few decimal places for 5-6 nodes, so I used `round()`. I don't think if this is happening bcoz of `get_chunk`. I'm looking into this. Let me know if you know why this might be happening.

dschult: That is almost certainly because of round-off error when computing/storing the numbers. Floating point values (double precision) have round-off errors typically around 10^-16, and so the rounding to 15 decimals aligns with that. In fact round-off error often becomes larger than that over more than one arithmetic operation, so it is nice that these errors are as small as they are. (round-off is different for different machines, for different orders of the operations, etc.. They aren't random, but they aren't predictable. :)

dschult: Another way to check that floating point numbers are close is [`math.isclose` ](https://docs.python.org/3/library/math.html#math.isclose) or the numpy version [`numpy.isclose`](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html).  They handle the edge cases where rounding gets tricky (relative error vs absolute error, etc).

dschult:
This PR has style issues. Next time you get to it, I think it is just one line with a space at the end of the line.

dschult:
It might be assuming too much to assert that the time should be less when chunking is used.
Of course, asserting that could allow you to track down *why* it is slower, I suppose. 

But should it always be slower? How might get_chunks slow it down?

Schefflera-Arboricola:
> It might be assuming too much to assert that the time should be less when chunking is used. Of course, asserting that could allow you to track down _why_ it is slower, I suppose.
> 
> But should it always be slower? How might get_chunks slow it down?

here, I have compared the time for running the parallel implementation of `betweenness_centrality` with the default chunking and with a custom chunking(done based on the `edge_betweenness_centrality` values). And I locally ran it on 8 cores and in github workflow test it runs on 2 cores, and for some machines, this test is failing and for some it is passing. 
Screenshot 2024-03-13 at 6 58 08 PM

This could be a nice example of why `get_chunk` might be useful as an option, that would let a user do custom chunking when they have a really big graph. But, I don't think we should put it as a test because it takes a lot of time and it also doesn't pass for all machines.

dschult:
That makes good sense to me.
Timing is dependent on so many things...  Hard to do well... 
Nice to have parameters to tweak for each situation and get_chunk is a good one to make available. :)

dschult (approved):
I think this is ready to be merged. So I'm approving it.

Are there any more changes planned for this PR?

Schefflera-Arboricola:
> Are there any more changes planned for this PR?
No, I think it's ready

Thanks @dschult !

PR #33, opened 2024-01-17
ENH : adding parallel implementations of `all_pairs_` algos

Schefflera-Arboricola:

Added the following :

- `all_pairs_dijkstra`
- `all_pairs_dijkstra_path_length`
- `all_pairs_dijkstra_path`
- `all_pairs_bellman_ford_path_length`
- `all_pairs_shortest_path`
- `all_pairs_shortest_path_length` - no speedups(the standard sequential implementation is already really fast, and even when the graph size was increased till 4000 nodes, there were negligible speedups, and it only took a minute or 2 to generate its heatmap)
- `all_pairs_node_connectivity`<sup>1</sup>
- `approximate_all_pairs_node_connectivity`<sup>1</sup>
- `all_pairs_all_shortest_paths`

<br>
<br>

[1] here, default chunking means having chunks of at most 10 nodes, because otherwise, it was taking too long with bigger node_chunks; and I also reduced the number of nodes while timing, but the speedups are pretty consistent(i.e. around 3). This heatmap alone took 2.5 hrs, and ran successfully after several tries. Also no heatmap for `approximate_all_pairs_node_connectivity` because for a random graph, it will result in the same heatmap as `all_pairs_node_connectivity`


nx_parallel/algorithms/shortest_paths/weighted.py
```diff
@@ -1,57 +1,82 @@
...
+    "all_pairs_bellman_ford_path_length",
+    "all_pairs_bellman_ford_path",
+]
+
+
+def all_pairs_dijkstra(G, cutoff=None, weight="weight"):
+    """Find shortest weighted paths and lengths between all nodes."""
+
+    def _calculate_all_pairs_dijkstra_subset(n):
+        return (n, single_source_dijkstra(G, n, cutoff=cutoff, weight=weight))
```
dschult, line 30: I think this should be `G.graph_object` passed into `single_source_dijkstra`. But maybe I'm not understanding...
```python
        return (n, single_source_dijkstra(G.graph_object, n, cutoff=cutoff, weight=weight))
```

nx_parallel/algorithms/shortest_paths/weighted.py
```diff
@@ -1,8 +1,87 @@
...
+
+def all_pairs_dijkstra_path(G, cutoff=None, weight="weight"):
+    """Compute shortest paths between all nodes in a weighted graph."""
+    if hasattr(G, "graph_object"):
+        G = G.graph_object
+
+    path = single_source_dijkstra_path
+
+    def _calculate_all_pairs_dijkstra_path_subset(n):
+        return (n, path(G, n, cutoff=cutoff, weight=weight))
```
dschult, line 63: Same comments in dijkstra as above regarding namespace for the function name and input args when they don't depend on `n`.

nx_parallel/algorithms/approximation/connectivity.py
```diff
@@ -0,0 +1,51 @@
...
+]
+
+
+def all_pairs_node_connectivity(G, nbunch=None, cutoff=None):
+    def _calculate_all_pairs_node_connectivity_subset(chunk):
+        for u, v in chunk:
+            k = local_node_connectivity(G, u, v, cutoff=cutoff)
+            all_pairs[u][v] = k
+            if not directed:
+                all_pairs[v][u] = k
```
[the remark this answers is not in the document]
Schefflera-Arboricola: > It might be worth defining this function based on the value of directed

here, I think you mean having 2 helper functions(one for when `directed` is True and one for when it's False), but i think passing `directed` as an argument in one function might be better.

Either way, I think the impact on the performance would be negligible but passing `directed` as an argument in the inside function might decrease the number of lookups in the outer namespace, so I think it would be better than having 2 different functions as that would reduce the readability of the code. Let me know what you think.

nx_parallel/algorithms/connectivity/connectivity.py
```diff
@@ -0,0 +1,80 @@
...
+from joblib import Parallel, delayed
+import nx_parallel as nxp
+
+__all__ = [
+    "all_pairs_node_connectivity",
+]
+
+
+def all_pairs_node_connectivity(G, nbunch=None, flow_func=None, get_chunks="chunks"):
+    """The parallel implementation first divides the a list of all permutation (in case
```
dschult, line 18:
```suggestion
    """The parallel implementation first divides a list of all permutation (in case
```

Schefflera-Arboricola:
`Tests / test (macos-latest, 3.11)` is running for more than 4 hrs. It is stuck at:
```
algorithms/centrality/tests/test_group.py ........................       [ 50%]
```

dschult:
That happened to me yesterday too.  I don’ think it has anything to do with the backend, but I’m not sure.

dschult:
Where does this PR stand? Is there more to do here?
Do you feel the improvements here are what you expect -- or are there other questions/issues you want to explore with this PR?

Schefflera-Arboricola:
I think we can merge this. The only algo not showing any speedups is `all_pairs_shortest_path_length` and its sequential implementation is already really fast. But, we should think about how to handle such functions in the whole of nx-parallel, i.e the functions that don’t show any speedups for the general random graphs we use in the current timing script and/or whose sequential networkx implementation is already really fast(like `number_of_isolates`, `all_pairs_shortest_path_length`, `global_reaching_centrality` and `local_reaching_centrality`(https://github.com/networkx/nx-parallel/pull/44), etc.). For some such algos setting the minimum chunk size to be really large might help(this didn't work for `all_pairs_shortest_path_length`), but then the input graph should also be really big. Just a thought, but having something like should_run = True only when the graph is really big might be helpful.

dschult (approved):

PR #34, opened 2024-01-18
ENH : added `square_clustering`

Schefflera-Arboricola:
**Please do "Squash and merge" while merging this PR**

PR #41, opened 2024-01-30
 updated `ParallelGraph` class

Schefflera-Arboricola:
these are needed in PR https://github.com/networkx/nx-parallel/pull/37 , https://github.com/networkx/nx-parallel/pull/34 and https://github.com/networkx/nx-parallel/pull/33 , so it's better to have them in a separate PR than in all of them.


dschult:
Are these needed for the ParallelGraph? We can use the underlying `G.graph_object` right?
But maybe I am missing a way in which the backend machinery needs to know e.g. the number of nodes.

PR #43, opened 2024-02-06
MAINT : added `default_benchmark_timeout` to `asv.conf.json`

Schefflera-Arboricola:
For 800 node graphs, the benchmarks fail and give a timeout error(60.0 s).

### DIFF
--- a/benchmarks/asv.conf.json
+++ b/benchmarks/asv.conf.json
@@ -16,5 +16,6 @@
     "env_dir": "env",
     "results_dir": "results",
     "html_dir": "html",
-    "build_cache_size": 8
+    "build_cache_size": 8,
+    "default_benchmark_timeout": 1200
 }
--- a/benchmarks/asv.conf.json
+++ b/benchmarks/asv.conf.json
@@ -17,5 +17,5 @@
     "results_dir": "results",
     "html_dir": "html",
     "build_cache_size": 8,
-    "default_benchmark_timeout": 1200
+    "default_benchmark_timeout": 1200,
 }


dschult:
I couldn't find `default_benchmark_timeout` in [the docs for asv](https://asv.readthedocs.io/en/latest/index.html).
Where did you find that?    it's a nice thing to be able to set.   :)

Schefflera-Arboricola:
> I couldn't find default_benchmark_timeout in [the docs for asv](https://asv.readthedocs.io/en/latest/index.html).
Where did you find that? it's a nice thing to be able to set. :)

here https://github.com/airspeed-velocity/asv/pull/1308
Also after increasing the default timeout here https://github.com/networkx/nx-parallel/pull/42 i was no longer getting the timeout error.

benchmarks/asv.conf.json
```diff
@@ -16,5 +16,6 @@
     "env_dir": "env",
     "results_dir": "results",
     "html_dir": "html",
-    "build_cache_size": 8
+    "build_cache_size": 8,
+    "default_benchmark_timeout": 1200
```
dschult, line 20:
```suggestion
    "default_benchmark_timeout": 1200,
```
You can put a comma at the end of the list... Python is ok with that. And it means you don't have to change that line when you add the inevitable next key/value pair to the dict. Same thing works for lists and tuples.

dschult (approved):
I approve this PR.
I make a comment below, but it doesn't need to be changed here unless you want to.

PR #48, opened 2024-02-20
MAINT : removed `NETWORKX_GRAPH_CONVERT`

Schefflera-Arboricola:
ref. https://github.com/networkx/networkx/pull/6840


dschult (approved):
Good catch!

Issue #52, opened 2024-03-11
RuntimeWarning when loading parallel backend


dschult:
I get this locally too, but the pytest logs on CI actions also show a warning:
```
  C:\Miniconda\envs\testing\lib\site-packages\networkx\utils\backends.py:143: RuntimeWarning: Error encountered when loading info for backend parallel: partially initialized module 'networkx' has no attribute '_dispatchable' (most likely due to a circular import)
```
I suspect nx-parallel is being imported as part of the networkx import and nx-parallel is importing networkx as part of its import, so we're getting a circular import.  

I think the solution is to delay importing networkx from within nx-parallel until after nx-parallel is fully imported. I haven't looked for all the places that import networkx but I suspect it is quite a few. 

Is there an easier approach? How does graphblas-algorithms do it? @Schefflera-Arboricola are you getting this warning also?

eriknw:
Other backends only get loaded when needed such as when calling one of their backends. The `networkx.backend_info` entry-point is used to provide information _without_ importing the backend (or networkx).

See:
- https://github.com/rapidsai/cugraph/blob/branch-24.04/python/nx-cugraph/_nx_cugraph/__init__.py
- https://github.com/python-graphblas/graphblas-algorithms/blob/main/_nx_graphblas/__init__.py

PR #53, opened 2024-03-15
BUG: moved `get_info` from `backend.py` to `_nx_parallel`

Schefflera-Arboricola:

fixes https://github.com/networkx/nx-parallel/issues/52

for ref. https://github.com/networkx/nx-parallel/pull/27


dschult:
I don;t think you need to make the additional parameters dict.  It is a helpful structure if you already know the parameter names and corresponding docs.  But when you are pulling from existing doc_strings on these functions, it is probably easier to just extend the `additional_docs` string with the parameter descriptions.

That is: you can add the parameter section of the doc_strings to the `additional_docs` field (leaving `additional_parameters blank) rather than parsing and making the dict structure required for `additional_parameters`.

dschult (approved):
This looks good to me.  Let’s try it.

PR #68, opened 2024-06-05
DRAFT: Adding config

Schefflera-Arboricola:
```.py
>>> import networkx as nx
>>> import nx_parallel as nxp
>>> nx.config
NetworkXConfig(backend_priority=['parallel'], backends=Config(parallel=NxpConfig(n_jobs=None, backend=None, return_as='list', verbose=0, timeout=None, pre_dispatch='2 * n_jobs', batch_size='auto', temp_folder=None, max_nbytes='1M', mmap_mode='r', prefer=None, require=None)), cache_converted_graphs=False)
>>> nx.config["backends"]["parallel"]["n_jobs"] = 4
>>> nx.config
NetworkXConfig(backend_priority=['parallel'], backends=Config(parallel=NxpConfig(n_jobs=4, backend=None, return_as='list', verbose=0, timeout=None, pre_dispatch='2 * n_jobs', batch_size='auto', temp_folder=None, max_nbytes='1M', mmap_mode='r', prefer=None, require=None)), cache_converted_graphs=False)
>>> configs = nxp.get_configs()
>>> configs
{'n_jobs': 4, 'backend': None, 'return_as': 'list', 'verbose': 0, 'timeout': None, 'pre_dispatch': '2 * n_jobs', 'batch_size': 'auto', 'temp_folder': None, 'max_nbytes': '1M', 'mmap_mode': 'r', 'prefer': None, 'require': None}
>>> configs["n_jobs"]
4
>>> G = nx.fast_gnp_random_graph(500, 0.8)
>>> H = nxp.ParallelGraph(G)
>>> print(nx.square_clustering(H))
```

### DIFF
--- a/nx_parallel/algorithms/cluster.py
+++ b/nx_parallel/algorithms/cluster.py
@@ -47,7 +47,8 @@ def square_clustering(G, nodes=None, get_chunks="chunks"):
     else:
         node_iter = list(G.nbunch_iter(nodes))
 
-    total_cores = nxp.cpu_count()
+    configs = nxp.get_configs()
+    total_cores = configs["n_jobs"]
 
     if get_chunks == "chunks":
         num_in_chunk = max(len(node_iter) // total_cores, 1)
@@ -55,7 +56,7 @@ def square_clustering(G, nodes=None, get_chunks="chunks"):
     else:
         node_iter_chunks = get_chunks(node_iter)
 
-    result = Parallel(n_jobs=total_cores)(
+    result = Parallel(**configs)(
         delayed(_compute_clustering_chunk)(node_iter_chunk)
         for node_iter_chunk in node_iter_chunks
     )
--- a/nx_parallel/utils/chunk.py
+++ b/nx_parallel/utils/chunk.py
@@ -1,6 +1,7 @@
 import itertools
 import os
 import networkx as nx
+import nx_parallel as nxp
 
 __all__ = ["chunks", "cpu_count", "create_iterables"]
 
@@ -20,7 +21,8 @@ def cpu_count():
     # Check if we are running under pytest
     if "PYTEST_CURRENT_TEST" in os.environ:
         return 2
-    return os.cpu_count()
+    else:
+        return nxp.get_configs("n_jobs")
 
 
 def create_iterables(G, iterator, n_cores, list_of_iterator=None):
--- a/nx_parallel/utils/config.py
+++ b/nx_parallel/utils/config.py
@@ -1,0 +1,48 @@
+from networkx.utils.configs import Config
+import joblib
+import inspect
+import networkx as nx
+from typing import Union
+
+__all__ = ["NxpConfig", "get_configs"]
+
+
+class NxpConfig(Config):
+    n_jobs: int = None
+    backend: str = None
+    return_as: str = "list"
+    verbose: int = 0
+    timeout: float = None
+    pre_dispatch: str = "2 * n_jobs"
+    batch_size: int = "auto"
+    temp_folder: str = None
+    max_nbytes: Union[int, str] = "1M"
+    mmap_mode: str = "r"
+    prefer: str = None
+    require: str = None
+
+    """
+    todo: use something like this instead:
+    for param_name, param in parallel_params.items():
+        if param_name not in ["self", "args", "kwargs"]:  # Exclude non-configurable params
+            param_name = param.default
+    """
+
+
+def get_configs(config=None):
+    parallel_params = inspect.signature(joblib.Parallel).parameters
+    configs = {
+        k: nx.config["backends"]["parallel"][str(k)] for k, v in parallel_params.items()
+    }
+    if config is None:
+        return configs
+    elif isinstance(config, list):
+        new_configs = {k: configs[k] for k in config if k in configs}
+        return new_configs
+    return configs[config]
+
+
+nx.config.backend_priority = [
+    "parallel",
+]
+nx.config["backends"]["parallel"] = NxpConfig()
--- a/.gitignore
+++ b/.gitignore
@@ -109,6 +109,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+nxp-dev/
 
 # Spyder project settings
 .spyderproject
--- a/nx_parallel/utils/__init__.py
+++ b/nx_parallel/utils/__init__.py
@@ -1,1 +1,2 @@
 from .chunk import *
+from .config import *


nx_parallel/utils/chunk.py
```diff
@@ -20,7 +21,8 @@ def cpu_count():
     # Check if we are running under pytest
```
Schefflera-Arboricola, line 21: todo: remove cpu_count and instead have the get_configs() and a function that returns default configs for testing.

nx_parallel/utils/config.py
```diff
@@ -0,0 +1,48 @@
...
+    elif isinstance(config, list):
+        new_configs = {k: configs[k] for k in config if k in configs}
+        return new_configs
+    return configs[config]
+
+
+nx.config.backend_priority = [
+    "parallel",
+]
+nx.config["backends"]["parallel"] = NxpConfig()
```
Schefflera-Arboricola, line 48: It won't be like this in future.

nx_parallel/algorithms/cluster.py
```diff
@@ -47,15 +47,18 @@ def _compute_clustering_chunk(node_iter_chunk):
...
+    configs["n_jobs"] = total_cores
 
     if get_chunks == "chunks":
         num_in_chunk = max(len(node_iter) // total_cores, 1)
         node_iter_chunks = nxp.chunks(node_iter, num_in_chunk)
     else:
         node_iter_chunks = get_chunks(node_iter)
 
-    result = Parallel(n_jobs=total_cores)(
+    result = Parallel(**configs)(
```
Schefflera-Arboricola, line 61: not sure if `Parallel` should be used here or `parallel_config`...

or something like this(like in PR#https://github.com/networkx/nx-parallel/pull/7)
```.py
with joblib.parallel_backend('<parallel_backend_name>'):
    return joblib.Parallel(n_jobs=-1)(calls)
```

_nx_parallel/update_get_info.py
```diff
@@ -58,7 +59,9 @@ def extract_docstrings_from_file(file_path):
                     and node.targets[0].id == "__all__"
                 ):
                     all_list = [
-                        expr.s for expr in node.value.elts if isinstance(expr, ast.Str)
+                        expr.value
+                        for expr in node.value.elts
+                        if isinstance(expr, ast.Constant)
```
Schefflera-Arboricola, line 64: Addressing deprecation warnings

```
nx-parallel/_nx_parallel/update_get_info.py:62: DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
  expr.s for expr in node.value.elts if isinstance(expr, ast.Str)
nx-parallel/_nx_parallel/update_get_info.py:62: DeprecationWarning: Attribute s is deprecated and will be removed in Python 3.14; use value instead
  expr.s for expr in node.value.elts if isinstance(expr, ast.Str)
```

nx_parallel/algorithms/cluster.py
```diff
@@ -47,15 +47,16 @@ def _compute_clustering_chunk(node_iter_chunk):
     else:
         node_iter = list(G.nbunch_iter(nodes))
 
-    total_cores = nxp.cpu_count()
+    n_jobs = nxp.cpu_count()
+    n_jobs = 1 if n_jobs is None else n_jobs  # changing the default
```
dschult, line 51: Let's try putting this into the central function as you described in the zoom call.

nx_parallel/algorithms/cluster.py
Schefflera-Arboricola: `parallel_backend` is a deprecated functionalities; used `parallel_config` here.

Schefflera-Arboricola:
For proper integration, I will need to change a few lines of code in all the algorithms. So far, I've modified `square_clustering` to work well with NetworkX's config, which included renaming `cpu_count` to `get_n_jobs`. However, I've reverted `get_n_jobs` back to `cpu_count` because the tests were failing, as other algorithms still use `cpu_count`. I'll proceed with updating the other algorithms once I receive the first review. I think changing the name across all algorithms right now will over-crowd this PR and will make it challenging to review. LMK if that seems good to you @dschult .

Thank you :)

Config.md
```diff
@@ -0,0 +1,48 @@
...
+>>> nxp_config = nx.config.backends.parallel
+>>> nxp_config.backend = "loky" # global config
+>>>
+>>> G = nx.complete_graph(20)
+>>> # using context manager
+>>> with nxp_config(n_jobs=4): # backend -> loky , n_jobs -> 4 , verbose -> 0
+...     nx.square_clustering(G, backend="parallel")
+...     nxp_config
+...     print()
+...     with nxp_config(n_jobs=3, verbose=15): # backend -> loky , n_jobs -> 3 , verbose -> 15
```
dschult, line 19: Why are the context manager calls set within the previous context manager call? Shouldn't they be separate -- in the sense that we end one context manager before starting the next configuration setting -- that is, should we dedent these inner context manager calls?

(and can we put the comments on the previous line so they don't make the line so long?

_nx_parallel/config.py
```diff
@@ -0,0 +1,25 @@
+from networkx.utils.configs import Config
+from typing import Union
+from dataclasses import dataclass, field
+
+__all__ = [
+    "ParallelConfig",
+    "_configs",
```
dschult, line 7: When do we choose to use the plural name `configs` and when do we use `config`? It seems like most of the code here and in networkx we use the singular (except for the module name in networkx which kind of makes sense because that code allows many different config to co-exist and the module name does need to be different from local names).  Should this be `_config`?  

Also, we don't usually include private names in `__all__`.  Should this be a public name instead of a private name?  It looks to me that this is only used to set the default config. So I would suggest keeping it private, naming it `_default_config`, and removing it from `__all__`. 

I'm actually a little worried that `ParallelConfig` is a public class. Do we want users to be creating their own config instances? I guess I don't understand the config paradigm fully.

nx_parallel/algorithms/cluster.py
```diff
@@ -47,18 +47,20 @@ def _compute_clustering_chunk(node_iter_chunk):
     else:
         node_iter = list(G.nbunch_iter(nodes))
 
-    total_cores = nxp.cpu_count()
+    configs = nxp.get_configs()
```
dschult, line 50: plural vs singular?

nx_parallel/utils/chunk.py
```diff
@@ -51,3 +65,27 @@ def create_iterables(G, iterator, n_cores, list_of_iterator=None):
...
         raise ValueError("Invalid iterator type.")
+
+
+def _get_configs(configs):
+    config_dict = asdict(configs)
+    config_dict.update(config_dict["backend_params"])
+    del config_dict["backend_params"]
+    config_dict["n_jobs"] = cpu_count(config_dict["n_jobs"])
+    if "PYTEST_CURRENT_TEST" in os.environ:
+        config_dict["backend"] = "loky"
```
dschult, line 76: Can you help me understand the `PYTEST_CURRENT_TEST` environment variable?

nx_parallel/utils/chunk.py
```diff
@@ -51,3 +65,27 @@ def create_iterables(G, iterator, n_cores, list_of_iterator=None):
...
+
+
+def get_configs(config=None):
+    """Returns the current configuration settings for nx_parallel."""
+    config_dict = _get_configs(nx.config.backends.parallel)
+    if config is None:
+        return config_dict
+    elif isinstance(config, list):
+        new_config = {k: config_dict[k] for k in config if k in config_dict}
+        return new_config
```
dschult, line 87: Is handling a list of config attribute names standard config treatment?

Config.md
Schefflera-Arboricola: here, I wanted to show how context managers inherit from the outer(or previous) context manager. And how we can keep making configs more and more specific by "(sort-of)nesting" context managers. LMK what you think. I think we can add another context manager that is independent of all these.

nx_parallel/utils/chunk.py
Schefflera-Arboricola: here i have just used it to check if we are running pytests, so that i can set the backend to "loky" otherwise with backend=None(default) we get errors from the `joblib.Parallel` call.

ref. https://docs.pytest.org/en/7.1.x/example/simple.html#pytest-current-test-env

nx_parallel/utils/chunk.py
Schefflera-Arboricola: I don't think so. But, it seemed helpful to me. I'll look into it more.

Config.md
dschult: If that is what we are showing, then let's put some words saying that it is one of the points/features of these examples.

nx_parallel/utils/chunk.py
dschult: hmmm....   It feels like we are changing the code when we are testing -- and that makes me uncomfortable. What would happen if someone ran their code with `backend=None`/default? Would that also get errors? I get that this makes it work, but it seems like we should be able to avoid this by setting a default of 'loky' for all ways that call with backend=None. 
:}/

dschult:
After reading [the blog about adding config](https://github.com/Schefflera-Arboricola/blogs/blob/main/networkx/GSoC24/Week11-12.md) I am more in favor of avoiding the use of the networkx config system for controlling the nx-parallel use of joblib. Said another way, I think it would be better to keep the user in the joblib config system and support no joblib related parameters in `nx.config`. 

Duplicating the joblib parameters doesn't make sense to me. And maintaining two (or creating a third) config system seems to ask for confusion and duplication.  Can we just rely on users using the joblib config system?

Schefflera-Arboricola:
> After reading [the blog about adding config](https://github.com/Schefflera-Arboricola/blogs/blob/main/networkx/GSoC24/Week11-12.md) I am more in favor of avoiding the use of the networkx config system for controlling the nx-parallel use of joblib. Said another way, I think it would be better to keep the user in the joblib config system and support no joblib related parameters in `nx.config`.
> 
> Duplicating the joblib parameters doesn't make sense to me. And maintaining two (or creating a third) config system seems to ask for confusion and duplication. Can we just rely on users using the joblib config system?

nx-parallel is a networkx backend that uses joblib(right now), not a joblib backend for graphs algorithms. So, it makes sense that we recommend our users to use the networkx's config context manager over the joblib's. Because, if a networkx user wants to use nx-parallel and with some other backend(s)(like, with nx-cugraphs -- like `backend_priority=[cugraph, parallel]`) then they would have to set the nx-parallel configs in the `nx.config` because this user would not want to run their code with a joblib's context manager when running with nx-cugraphs' implementation. And doing what you are suggesting will deprive nx-parallel users of all those features and functionalities that lets(or will let) them use multiple networkx backends together and this would probably exclude nx-parallel from the networkx backend ecosystem.

And we only have one config system i.e. NetworkX's config, which can also be used as a context manager. And joblib's context manager will not work as expected with the config system implemented in this PR(as outlined in the blog you mentioned above).

The only maintenance needed will be to update the `ParallelConfig` class, if joblib ever changes the args of the `parallel_config` class, which I think is very unlikely to happen very frequently. Also if it does happen, joblib has an appropriately long deprecation period. 

Also, the main concern of mine(also mentioned in the blog) with the current implementation in this PR, is that the fetching of the global configs in every function call is redundant but it is not very expensive. It's like converting `ParallelGraph` instance into a `nx.Graph` instance in every nx-parallel algorithm.

Thank you for your comments. I hope I addressed them all.

nx_parallel/utils/chunk.py
Schefflera-Arboricola: > hmmm.... It feels like we are changing the code when we are testing -- and that makes me uncomfortable. 

We were doing this earlier also. We were changing `n_jobs=2` for when we were testing.

> What would happen if someone ran their code with `backend=None`/default? Would that also get errors? 

As mentioned in the previous comment, we do get the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/aditijuneja/Desktop/nx-parallel/nxp-dev/lib/python3.12/site-packages/networkx/utils/decorators.py", line 789, in func
    return argmap._lazy_compile(__wrapper)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<class 'networkx.utils.decorators.argmap'> compilation 20", line 3, in argmap_square_clustering_17
  File "/Users/aditijuneja/Desktop/nx-parallel/nxp-dev/lib/python3.12/site-packages/networkx/utils/backends.py", line 910, in __call__
    return self._convert_and_call(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aditijuneja/Desktop/nx-parallel/nxp-dev/lib/python3.12/site-packages/networkx/utils/backends.py", line 1322, in _convert_and_call
    result = getattr(backend, self.name)(*converted_args, **converted_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aditijuneja/Desktop/nx-parallel/nx_parallel/algorithms/cluster.py", line 59, in square_clustering
    with parallel_config(**config):
         ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aditijuneja/Desktop/nx-parallel/nxp-dev/lib/python3.12/site-packages/joblib/parallel.py", line 375, in __init__
    backend = self._check_backend(
              ^^^^^^^^^^^^^^^^^^^^
  File "/Users/aditijuneja/Desktop/nx-parallel/nxp-dev/lib/python3.12/site-packages/joblib/parallel.py", line 439, in _check_backend
    if backend.nesting_level is None:
       ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'nesting_level'
```

> I get that this makes it work, but it seems like we should be able to avoid this by setting a default of 'loky' for all ways that call with backend=None.

I have changed the default backend to `loky` so we would not have to change it from `None` to `"loky"` at the time of testing. Also, now a nx-parallel user will also not get the above error if they don't specify a backend. And that is what happens when a user doesn't specify any backend in the joblib's `parallel_config` context manager because the default backend for the `joblib.Parallel` is `loky`. I'll make sure to document this in the `Config.md` clearly.

dschult:
> nx-parallel is a networkx backend that uses joblib(right now), not a joblib backend for graphs algorithms. So, it makes sense that we recommend our users to use the networkx's config context manager over the joblib's. Because, if a networkx user wants to use nx-parallel with some other backend(s)(like, with nx-cugraphs -- like backend_priority=[cugraph, parallel]) then they would have to set the nx-parallel's configs in the nx.config because this user would not want to run their code with a joblib's context manager when running with nx-cugraphs' implementation(because it will probably fail). And doing what you are suggesting will deprive nx-parallel users of all those features and functionalities that lets(or will let) them use multiple networkx backends together and this would probably exclude nx-parallel from the networkx backend ecosystem.

Can you help me understand this better? Are we expecting every networkx backend to support running in an environment with every other networkx backend's configs? Said another way, why do we need to worry about users having two nx backends in their priority order when making nx-parallel look for joblib config values? It seems like the nx-parallel backend will look for joblib config settings and the nx-cugraph backend will look for cugraph config settings. How would these break each other?

Schefflera-Arboricola:
> Can you help me understand this better? Are we expecting every networkx backend to support running in an environment with every other networkx backend's configs? Said another way, why do we need to worry about users having two nx backends in their priority order when making nx-parallel look for joblib config values? It seems like the nx-parallel backend will look for joblib config settings and the nx-cugraph backend will look for cugraph config settings. How would these break each other?

It looks like I should expand more on what I wrote in blog about how NetworkX's config works and how `nx-parallel` integrates with the NetworkX configuration system and why it's essential to have its configs in `nx.config` instead of relying solely on joblib's context manager.

The unified configuration management system within NetworkX (`nx.config`) ensures consistency and ease of use when dealing with different backends. Each backend, such as `nx-parallel` and `nx-cugraph`, will have its configurations managed within this unified system.

If each backend looked for its configurations separately (e.g., `nx-parallel` using joblib's context manager and `nx-cugraph` using its own system), it could lead to conflicts and difficulties in managing multiple backends. By having all backend configurations within `nx.config`, we ensure that they can coexist and be prioritized without breaking each other. Also this is easier for the user also to specify all the configs for all the backends at one place.

### Addressing your questions

- **Expectations of Backend Configurations:**
  
  We do not expect every NetworkX backend to support every other backend’s configurations. Instead, we expect them to adhere to the unified configuration system provided by `nx.config`. By not making `nx-parallel` look for joblib's context manager, we maintain a consistent configuration environment within NetworkX. And I think compatibility with networkx is more important than joblib because as quoted above :
   
  > nx-parallel is a networkx backend that uses joblib(right now), not a joblib backend for graphs algorithms. 

- **Importance of `nx.config` for `nx-parallel`:**

  Joblib configurations are part of `nx-parallel` configurations. To use `nx-parallel` effectively with other backends, these configurations need to be included within `nx.config`. Example:
  
  ```python
  NetworkXConfig(backend_priority=[cugraph, parallel], backends=Config(cugraph=<configs_for_cugraphs>, parallel=ParallelConfig(backend='loky', n_jobs=None, verbose=0, temp_folder=None, max_nbytes='1M', mmap_mode='r', prefer=None, require=None, inner_max_num_threads=None, backend_params={})), cache_converted_graphs=True)
  ```
  
  This example shows how `nx.config` can manage multiple backends, including `nx-parallel`, ensuring that all configurations are in one place and can be managed consistently.


So, to summarise, keeping all backend configs within `nx.config` ensures consistency and ease of management, prevents conflicts that could arise from using multiple backends with separate configuration systems. By having `nx-parallel` configurations within `nx.config`, we ensure that it can work seamlessly with other NetworkX backends. And that's why we use `nx.config` over joblib's context manager. Also joblib's context manager cannot fix global configs like `nx.config.backends.parallel.n_jobs = 8`, it can only set configs in a context.

If you need further clarification or have additional questions, please let me know.

dschult:
Thanks for that description of the advantages of using nx.config.backends over relying on other configuration systems.  I still have questions/concerns and I'll try to describe them here.

- if each backend relied on separate configuration systems, how might that lead to conflicts? Similarly, how would that lead to difficulties when using multiple backends? You would set up the config for joblib however you like. Then set up the config for cugraphs however you like, then set up the networkx backend_priority in whatever order you prefer. I don't see how conflicts could happen. And I don't see how it makes it simpler to put all the config options for all the backends in a single line of code. That seems more complicated to me -- not simpler.
- When we make all backends use the networkx config system, (as far as I can tell) the only consistency that is increased is consistency in the function names used to set/change parameters. People still have to learn the separate config systems to know what config options there are and what they do. So, they learn joblib and those parameters, then they learn cugraph and those parameters, and then they have to learn a third system for setting all those values.  Wouldn't it be easier if they just learn the config system for joblib and then cugraph and then the only config option that matters for nx.config.backends is the backend_priority. (and with the nx config number of parameters being small and easily explored it should take less time to learn the nx.config system than if it contains all the parameters for all the backends).
- In what way does the nx-parallel config system work seemlessly with the other backends when it is housed in nx-parallel? Isn't it independent of the other backend parameter configs? Wouldn't setting joblib parameters for nx-parallel work seemlessly with nx-cugraph? I guess this comes down to: what are the conflicts we are trying to avoid?
- If I understand correctly, joblib's context manager is only one way to set the config for joblib. Any of the existing methods for setting joblib parameters should work if we let users manage the joblib config parameters themselves. But maybe I'm missing something here. Also, setting n_jobs=8 in nx.config is not a global config setting either right? It doesn't impact joblib processes run outside of nx-parallel.  Do you mean "global within nx-parallel"?

Thanks for this discussion as I think I'm learning about config systems generally as part of it. :)

Schefflera-Arboricola:
> if each backend relied ..... complicated to me -- not simpler.

Sure, backends can have a different configs system if they want. But, in case of nx-parallel, if we only use joblib's config system, there are certain issues, like as mentioned above, a user won't be able to define global configs as joblib only offers context manager `parallel_config` to specify/update the configs, because for joblib these are not config options but parameters. So, then if the user only uses joblib’s context manager and puts all their code in a `joblib.parallel_config` context then while running the nx-cugraph implementation the joblib’s context manager will not be ignored and joblib will be used, which is not what a user would expect.

And one does not need to “put all the config options for all the backends in a single line of code”. These are the default configs:

```py
NetworkXConfig(
    backend_priority=[cugraph, parallel],
    backends=Config(
        cugraph=<configs_for_cugraphs>,
        parallel=ParallelConfig(
            backend='loky',
            n_jobs=None,
            verbose=0,
            temp_folder=None,
            max_nbytes='1M',
            mmap_mode='r',
            prefer=None,
            require=None,
            inner_max_num_threads=None,
            backend_params={}
        )
    ),
    cache_converted_graphs=True
)
```

And one can modify the ones they want like this:

```
nxp_configs = nx.config.backends.parallel
nxp_configs.backend = “threading”
nxp_configs.n_jobs = 7

nxc_configs = nx.config.backends.cugraph
…
```

I hope this looks less complicated to you.

> When we make all .... all the backends).

As mentioned in the Config.md, one can use the `nxp_configs`(from above) just like they use `joblib.parallel_config` context manger, so a user does not need to learn a third config system. Also, someone using a networkx backend is expected to understand networkx's config, I think, and then they would learn config systems specific to the backend(which right now should be very easy for the nx-parallel backend). And I prefer it this way, because nxp_configs can be used as a global variable as well as a context. 

> In what way does... we are trying to avoid?

I’m not sure if I understand this one properly... but most of the code for managing different backends and their configs is in the main networkx repo not in nx-parallel. And the conflict is that to set the joblib’s configs we need to use the joblib’s context manager which won’t be ignored in nx-cugraphs implementation and also we cannot set global configs in joblib.
 
> If I ....in nx-parallel"?

Yes, afaik joblib's context manager is the only way to set the config for joblib.Parallel. No, right now a user cannot use joblib’s context to set joblib’s parameters in nx-parallel. They can but it won’t work as expected. nx-parallel configs and joblib’s configs are independent but same. Also, in joblib n_jobs is a parameter not a config, so when I say ”globally“ or “global config” I mean it in nx-parallel not joblib.

Thank you :)

Schefflera-Arboricola:
Made default `n_jobs=-1` instead of `None` because I think this will be beneficial for the new users, as it would allow them to run their algorithms on all CPU cores(with the default `backend="loky"`) without needing to understand or configure the parallel options in detail. LMK what you think. Thanks!

Schefflera-Arboricola:
closing this PR. 
Approach in PR https://github.com/networkx/nx-parallel/pull/75 is more compelling.

PR #72, opened 2024-07-31
feat ✨ : Add closeness centrality and floyd warshall feature

Fre0Grella:
Hi, i'm marco, a students from the university of Bologna.
For my final project i want to contribute this repository adding the closeness centrality and the floyd warshall algorithms  for numpy.
The closeness centrality use the newly implemented floyd warshall.
The parallelized floyd warshall is a version of tiled FW referenced in the code.

### DIFF
--- a/nx_parallel/algorithms/centrality/closeness.py
+++ b/nx_parallel/algorithms/centrality/closeness.py
@@ -2,7 +2,7 @@ __all__ = ["closeness_centrality"]
 
 
 def closeness_centrality(
-    G, u=None, distance=None, wf_improved=True, get_chunks="chunks"
+    G, u=None, distance=None, wf_improved=True, blocking_factor=None
 ):
     """The parallel implementation first divide the nodes into chunks and
 
@@ -10,8 +10,8 @@ def closeness_centrality(
 
     Parameters
     ----------
-    get_chunks : str, function (default = "chunks")
-        A function that takes in a list of all the nodes as input and returns an
-        iterable `node_chunks`. The default chunking is done by slicing the
-        `nodes` into `n` chunks, where `n` is the number of CPU cores.
+    blocking_factor : number
+        The number used for divinding the adjacency matrix in sub-matrix.
+        The default blocking factor is get by finding the optimal value
+        for the core available
     """
--- a/nx_parallel/algorithms/centrality/closeness.py
+++ b/nx_parallel/algorithms/centrality/closeness.py
@@ -1,0 +1,17 @@
+__all__ = ["closeness_centrality"]
+
+
+def closeness_centrality(
+    G, u=None, distance=None, wf_improved=True, get_chunks="chunks"
+):
+    """The parallel implementation first divide the nodes into chunks and
+
+    networkx.closeness_centrality : https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.closeness_centrality.html
+
+    Parameters
+    ----------
+    get_chunks : str, function (default = "chunks")
+        A function that takes in a list of all the nodes as input and returns an
+        iterable `node_chunks`. The default chunking is done by slicing the
+        `nodes` into `n` chunks, where `n` is the number of CPU cores.
+    """
--- a/nx_parallel/algorithms/centrality/closeness.py
+++ b/nx_parallel/algorithms/centrality/closeness.py
@@ -1,10 +1,18 @@
+"""
+Closeness centrality measures.
+"""
+
+from joblib import Parallel, delayed
+import nx_parallel as nxp
+
 __all__ = ["closeness_centrality"]
 
 
 def closeness_centrality(
     G, u=None, distance=None, wf_improved=True, blocking_factor=None
 ):
-    """The parallel implementation first divide the nodes into chunks and
+    """The parallel implementation of closeness centrality, that use parallel tiled floy warshall to find the
+    geodesic distance of the node
 
     networkx.closeness_centrality : https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.closeness_centrality.html
 
@@ -15,3 +23,42 @@ def closeness_centrality(
         The default blocking factor is get by finding the optimal value
         for the core available
     """
+    if G.is_directed():
+        G = G.reverse()  # create a reversed graph view
+
+    A = nxp.floyd_warshall_numpy(G, blocking_factor=blocking_factor)
+
+    len_G = len(G)
+
+    closeness_dict = Parallel(n_jobs=-1)(
+        delayed(_closeness_measure(n, wf_improved, len_G))(A) for n in A
+    )
+
+    if u is not None:
+        return closeness_dict[u]
+    return closeness_dict
+
+
+def _closeness_measure(n, wf_improved, len_G):
+    """calculate the closeness centrality measure of one node using the row of edges i
+
+    Parameters
+    ----------
+    n : 1D numpy.ndarray
+        the array of distance from every other node
+
+    Returns
+    -------
+    k : numebr
+        the closeness value for the selected node
+    """
+    totsp = sum(n)
+    closeness_value = 0.0
+    if totsp > 0.0 and len_G > 1:
+        closeness_value = (len(n) - 1.0) / totsp
+        # normalize to number of nodes-1 in connected part
+        if wf_improved:
+            s = (len(n) - 1.0) / (len_G - 1)
+            closeness_value *= s
+
+    return closeness_value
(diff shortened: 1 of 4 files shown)


Schefflera-Arboricola:
@Fre0Grella nx-parallel algorithms receive a `ParallelGraph` object, so you need to convert it back to the `nx.Graph` object to run the usual graph object methods on it. That's why you'll see the following lines in all the nx-parallel algorithms

```py
        G = G.graph_object
```

Hopefully this will make the tests pass.

Thanks!

Fre0Grella:
@Schefflera-Arboricola Hi, thanks for the first advice, i hope you could help with this one too.
Running the test i got an error that i don't understand;
on the closeness.py file, when i use joblib Parallel to call the `_closeness_measure` function (that from a row of the adjacency matrix get the value of closeness for the node rappresented by the row) i get the error in the joblib code saying `TypeError: 'numpy.float64' object is not callable`.
I tried changing the data format thinking that joblib doesn't support numpy data structure but he kept saying in this case `TypeError: 'float' object is not callable`.

Schefflera-Arboricola:
Hi @Fre0Grella, Thank you for your contribution :)

Is this PR ready for review? I will aim to get to it as soon as possible, though I cannot provide a specific timeline at the moment. I'm also tagging @dschult in case he would like to review it.

Additionally, could you let us know if there is a deadline for your final project and if this PR needs to be merged before then?

Thank you :)

Fre0Grella:
Hi @Schefflera-Arboricola, the PR is ready for review although i aim to improve the performance further. The dead line of the project is due the 18 of September.
I doesn't need the PR to be merged before that

Schefflera-Arboricola:
Thanks @Fre0Grella! I think instead of `nxp.cpu_count` using `nxp.get_n_jobs` would have fixed the failing CI tests because we renamed it recently. If you still have the bandwidth please feel free to merge the main branch again. Really sorry for the delayed review!

Fre0Grella:
> Thanks @Fre0Grella! I think instead of `nxp.cpu_count` using `nxp.get_n_jobs` would have fixed the failing CI tests because we renamed it recently. If you still have the bandwidth please feel free to merge the main branch again. Really sorry for the delayed review!

Not sure if i'm doing something wrong but replacing `nxp.cpu_count` to `nxp.get_n_jobs` isn't working. On the test i get `AttributeError: module 'nx_parallel' has no attribute 'get_n_jobs'`

nx_parallel/algorithms/shortest_paths/dense.py
```diff
@@ -0,0 +1,250 @@
...
+                if block_i != primary_block or block_j != primary_block:
+                    i_range = _block_range(blocking_factor, block_i)
+                    j_range = _block_range(blocking_factor, block_j)
+                    if is_prime:
+                        if block_i == no_of_primary - 1:
+                            i_range = (block_i * blocking_factor, n - 1)
+                        if block_j == no_of_primary - 1:
+                            j_range = (block_j * blocking_factor, n - 1)
+                    params.append((i_range, j_range))
+        Parallel(n_jobs=(no_of_primary - 1) ** 2, require="sharedmem")(
```
Schefflera-Arboricola, line 91: Recently we have made it so that all the parameters of the `Parallel` call are controlled by the user and we don't add anything in it. I would highly encourage you(@Fre0Grella) to join our weekly nx-parallel meetings :  https://scientific-python.org/calendars/networkx.ics , if you can. I think your insights and feedback would really help us shape the nx-parallel project in a better way. Thank you :)

Schefflera-Arboricola:
All the tests are passing now and the style test should also pass once you update this PR branch with the recent fix that got merged. Thank you :)

Fre0Grella:
It would be awesome if we can merge this PR before 1st of october so i can present the project directly from the NetworkX repository to the exam commision. Thanks a lot for the support along the project. I'm also interest in further contribuition to the NetworkX project

Fre0Grella:
@Schefflera-Arboricola hey, the PR is ready to be merged, it only need a Review

nx_parallel/algorithms/shortest_paths/dense.py
```diff
@@ -0,0 +1,251 @@
...
+    results = [r for r in results if r is not None]
+
+    if len(results) <= 0:
+        # This check if a number is prime, although repeat process with a non prime number
+        best_divisor, _ = _find_nearest_divisor(x - 1, y)
+        return best_divisor, True
+    # Find the best divisor
+    best_divisor, _, _ = min(results, key=lambda x: x[2])
+
+    return best_divisor, False
```
dPys, line 180: Curious-- for large graphs with prime numbers of nodes, it seems like the final block could be disproportionately larger, leading to load imbalance across cores?

nx_parallel/algorithms/centrality/closeness.py
```diff
@@ -0,0 +1,73 @@
...
+    # print(totsp)
+    closeness_value = 0.0
+    if totsp > 0.0 and len_G > 1:
+        closeness_value = (len(n_reachable) - 1.0) / totsp
+        # normalize to number of nodes-1 in connected part
+        if wf_improved:
+            s = (len(n_reachable) - 1.0) / (len_G - 1)
+            closeness_value *= s
+
+    return k, closeness_value
```
dPys, line 73: I wonder if it could make sense to include an initial check in this function for graph disconnectedness, the goal being to "fail fast" to prevent edge-case scenarios where disconnected nodes lead to infinite geodesic distances. WDYT?

Fre0Grella: Yeah, a transitive closure test could be added in the future i think

nx_parallel/algorithms/shortest_paths/dense.py
dPys: Hm, what about just swapping:
```
best_divisor, _ = _find_nearest_divisor(x - 1, y)
```
with something as simple as:
```
best_divisor = max(1, x // y)
```
to distribute load approximately evenly if no divisor is found?

nx_parallel/algorithms/shortest_paths/dense.py
```diff
@@ -0,0 +1,251 @@
...
+def _calculate_divisor(i, x, y):
+    if x % i == 0:
+        divisor1 = result2 = i
+        result1 = divisor2 = x // i
+        # difference1 = abs((result1 - 1) ** 2 - y)
+        difference1 = abs(result1 - y)
+        # difference2 = abs((result2 - 1) ** 2 - y)
+        difference2 = abs(result2 - y)
+
+        if difference1 < difference2:
```
dPys, line 143: Could be missing something obvious, but aren't `divisor1` and `divisor2` actually going to be the same thing here (`i`)? i.e.  do we even need to compare difference1 with difference2?

I wonder whether this could be simplified to something like:
```
def _calculate_divisor(i, x, y):
    if x % i == 0:
        quotient = x // i
        return (i, quotient, abs(i - y))
    return None
```
?

Fre0Grella: Not really the same thing, the idea is to find the divisor of x ( size of matrix) most similar to y (core available).
To find i try to divide x to all number from 1 to sqrt(x) so when i do x // i, divisor1 is gonna be i, the result of x // i is gonna be divisor2.

An example to clarify:
x=10
y=4
i=2
 x // i = 5
So divisor1 = 2, divisor2 = 5

divisor2 is the most similar to 4 so i choose 5 as divisor

dPys: Thanks for the clarification on this @Fre0Grella 

I think what I'm still struggling with is the need for two separate comparisons. Since `divisor1` is `i` and `divisor2` is `x // i`, aren't you always comparing `i` and `x // i`? The current implementation is functionally correct, but this seems like it could be greatly simplified without losing the intent of matching `divisor2` more closely to `y`.

For example, maybe we could directly compute the abs difference of both `i` and `x // i` with `y`, then just return whichever is closest? And this would avoid the extra complexity in checking two sets of variables. So, building on my (incomplete) alternative suggestion previously, this might look something like:
```
def _calculate_divisor(i, x, y):
    if x % i == 0:
        divisor = x // i
        return (i, divisor, min(abs(i - y), abs(divisor - y)))
    return None
```
keeps the logic intact but then we don't need to assign and compare `divisor1` and `divisor2` explicitly. Let me know what you think / if I'm missing something!

dPys:
@Fre0Grella -- I should mention (since I haven't yet) that this is excellent work. @Schefflera-Arboricola-- hopefully after the remaining polishing we can get this thing merged?

PR #75, opened 2024-08-19
ENH: Adding and documenting configs in nx-parallel 

Schefflera-Arboricola:
This PR makes nx-parallel compatible with networkx's config system and documents how one can configure nx-parallel using `joblib.parallel_config`.

ref. PR https://github.com/networkx/nx-parallel/pull/68 for more context
ref. `Config.md` file in this PR for usage.

Thank you :)


Schefflera-Arboricola:
Notes for reviewers(@dschult):
- TODOs(not done yet to avoid over-crowding this PR; for ease of review; only done for `square_clustering` right now):
    - add the `_set_nx_config` decorator to all algorithms 
    - rename `cpu_count()` to `get_n_jobs()`
    - rename `total_cores` to `n_jobs`
- Please LMK if it would be possible to get this PR also reviewed and merged before GSoC ends(on August 26th) based on your schedule, so that I can accordingly mention it in my GSoC final report. I'm open to scheduling a meeting to discuss this PR if that would help you review this better and faster.

_nx_parallel/config.py
```diff
@@ -0,0 +1,57 @@
...
+    inner_max_num_threads: int = None
+    backend_params: dict = field(default_factory=dict)
+
+
+_config = ParallelConfig()
+
+
+def _set_nx_config():
+    """Decorator to set the configuration for the parallel computation
+    of the nx-parallel algorithms."""
```
dschult, line 35: I think I see why you put the `_set_nx_config` function here (the module is config.py afterall). But the decorator isn't used in the `_nx_parallel` package. It is used in `nx_parallel`. It feels like this function might be better found if it is in the main package (perhaps in utils?).

Also, the name sounds like we are setting the nx_config values. What do you think of `_allow_nx_config` or `_add_nx_config`?

Schefflera-Arboricola: I initially did have the `_set_nx_config` decorator in `utils` but then I was getting the following circular import error. So, then I moved it to `_nx_parallel`.

```
>>> import networkx as nx
>>> G = nx.path_graph(10)
>>> nx.square_clustering(G, backend="parallel")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/aditi/Desktop/nx-parallel/venv/lib/python3.12/site-packages/networkx/utils/decorators.py", line 789, in func
    return argmap._lazy_compile(__wrapper)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<class 'networkx.utils.decorators.argmap'> compilation 20", line 3, in argmap_square_clustering_17
  File "/Users/aditi/Desktop/nx-parallel/venv/lib/python3.12/site-packages/networkx/utils/backends.py", line 910, in __call__
    return self._convert_and_call(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aditi/Desktop/nx-parallel/venv/lib/python3.12/site-packages/networkx/utils/backends.py", line 1305, in _convert_and_call
    backend = _load_backend(backend_name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aditi/Desktop/nx-parallel/venv/lib/python3.12/site-packages/networkx/utils/backends.py", line 445, in _load_backend
    rv = _loaded_backends[backend_name] = backends[backend_name].load()
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/importlib/metadata/__init__.py", line 205, in load
    module = import_module(match.group('module'))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1310, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/Users/aditi/Desktop/nx-parallel/nx_parallel/__init__.py", line 1, in <module>
    from .algorithms import *
  File "/Users/aditi/Desktop/nx-parallel/nx_parallel/algorithms/__init__.py", line 13, in <module>
    from .cluster import *
  File "/Users/aditi/Desktop/nx-parallel/nx_parallel/algorithms/cluster.py", line 10, in <module>
    @nxp._set_nx_config()
     ^^^^^^^^^^^^^^^^^^
AttributeError: partially initialized module 'nx_parallel' has no attribute '_set_nx_config' (most likely due to a circular import)
>>> 
```

And yes the name does seem a bit off. Thank you for pointing it out. 
Also, I'm not in fair of using `allow_` or `add_` because the decorator first checks the value of the `nx_config` and then accordingly adds or not adds the networkx's config.

dschult: That makes sense...  The decorator has to be loaded into the namespace before it can be used. I think changing the order of imports inside `nx_parallel.__init__.py` to put `utils` first will fix the "partially initialized" problem (unless something inside `utils` uses the decorator). 

I think `allow_` indicates that the nx config won't always be used, but it can be used and without the decorator it can't be used. The term `set_` sounds like something is getting set (either nx_config values are getting set, or you are setting up the function to always use nx_config).  Maybe there is a better term?

Schefflera-Arboricola: Moving `utils` on top is giving the same error. 
I think (and I'm probably wrong) that it might have to do with this lazy compiling-- `argmap._lazy_compile(__wrapper)(*args, **kwargs)`. But, idk much about how the `argmap._lazy_compile` works.

> ... Maybe there is a better term?

does `_configure` or `_configure_context` sounds good to you?

dschult: Ahh...  try calling it with `@nxp.utils._set_config(`. Referring to the submodule gives it enough closer to allow the lookup to find _set_config. (it needs both the order change and the explicit module namespace.)

I like the `configure` word... but that decorator is not configuring. How about `_setup_dual_config`? hmm... maybe too long... but it's not *that* long...

Schefflera-Arboricola: It worked but I don't understand exactly how and why... I'd like to look into it; could pls direct me to where I should look to understand this?

Also `dual` gives a sense that 2 config systems are being setup, but it only sets up one. What do you think of `_apply_config` or `apply_parallel_config`?

dschult: Interesting  -- I thought it was setting up both parallel_config and nx_config? And then both are ready to be used...  

For where to look for info about importing namespaces, I don't know a place that talks about the mechanics of partially completed imports. `importlib` might be a place to look, but the docs seem to be more about how to use it than what goes on under the hood. I do know that importing a package runs it's `__init__.py` file. And that often imports other subpackages with their `__init__.py` files. Names can be added explicitly, or by using `*` in which case the subpackage's attribute `__all__` is used to figure out which names to add to the outer namespace.  But -- this issue involves code run by `__init__.py` that calls another part of nxp. First of all, that other part of nxp must have already been defined (that is, `utils` must have been imported before you can use it). What we ran into today is that it seems that the current/outer namespace is not fully accessible -- perhaps each layer only becomes accessible when that `__init__.py` file has finished, but I didn't know that restriction, so I am guessing as to why we can't use `nxp._set_config` as soon as `from utils import *` is finished. At the terminal I would expect to be able to use a function that was imported on one line as soon as that line completes. ???!?!  Anyway, by calling `nxp.utils._set_config` we only need to access the `utils` namespace -- instead of the main nxp namespace (which has not finished its initialization). The `utils` has finished its initialization. At least that is what I think is the reason we can access `nxp.utils._set_config`. 

One good way to reverse engineer the import system is to set up an example package with lots of print statements to see what gets run in which order. But I haven't done that yet... :)

_nx_parallel/__init__.py
```diff
@@ -13,6 +13,11 @@ def get_info():
         "short_summary": "Parallel backend for NetworkX algorithms",
         "default_config": _config,
         "functions": {
+            "_set_nx_config": {
+                "url": "https://github.com/networkx/nx-parallel/blob/main/nx_parallel/utils/decorators.py#L12",
+                "additional_docs": "Decorator to set the configuration for the parallel computation of the nx-parallel algorithms.",
+                "additional_parameters": None,
+            },
```
dschult, line 20: hmmm..  I am surprised that this shows up here because it is a private function.
Would it be better to have `get_info` only include the public functions?  The private functions should not be backends for networkx right?

Actually, I think the `get_info` probably shouldn't include any of the `utils` functions... Maybe just the functions that are backend functions for networkx...  But maybe I am asking too much of this system... :) Let's work on this/that in a later PR.

Schefflera-Arboricola: > hmmm.. I am surprised that this shows up here because it is a private function. Would it be better to have `get_info` only include the public functions? The private functions should not be backends for networkx right?

If by "private function" you mean that `_set_nx_config` is not in `__all__`, then that's not true, bcoz it is.

> Actually, I think the `get_info` probably shouldn't include any of the `utils` functions... Maybe just the functions that are backend functions for networkx... But maybe I am asking too much of this system... :) Let's work on this/that in a later PR.

Since none of the functions in networkx's utils folder have a `_dispatchable` decorator, I created PR https://github.com/networkx/nx-parallel/pull/78 that skips functions in `utils` while creating the `get_info` dict.

_nx_parallel/config.py
Schefflera-Arboricola: > Interesting -- I thought it was setting up both parallel_config and nx_config? And then both are ready to be used...

I renamed the `nx_config` configuration in `nx.config.backends.parallel` to `active`. I hope that make it more clear.

When `active=True` we use configs in `nx.config.backends.parallel`(i.e. put these configs in a `joblib.parallel_config` and make the function call in this context) and when it's `False` we don't do anything(just simply calls the function); the `joblib.Parallel` in nx-parallel algorithms itself looks for the configs set by the user in the `joblib.parallel_config`.

_nx_parallel/__init__.py
dschult: I was saying that `_set_nx_config` is private because it starts with an underscore (which seems to be a standard way to indicate a function that is not part of the public API). Perhaps we shouldn't be putting `_set_nx_config` into the `__all__` attribute.  I guess if we always call it as `@nxp.utils._set_nx_config` it doesn't need to be in `__all__` we just import it explicitly as part of `utils.__init__.py` with something like `from decorators import _set_nx_config`.  Then, even though utils is imported (via *) in the main `__init__.py`, my understanding is that `_set_nx_config` will not be added to the main namespace. But it would be worth checking that.

_nx_parallel/config.py
dschult: So, the user can use `parallel_config` at any time in any function. But this decorator makes it also possible to use nx_config (and either/both config systems). That's why I am thinking of it as enabling dual configs systems (nx_config and parallel_config). We can't turn off the joblib config system. So we are adding a new one, resulting in two config systems.

Schefflera-Arboricola: yes, we can't turn off the joblib config system, but we can overwrite it, and that is what is happening here. If someone enables networkx's config and then run the function in `joblib.parallel_config` like this:

```python
nx.config.backends.parallel.active = True
nx.config.backends.parallel.n_jobs = 6

with joblib.parallel_config(n_jobs=4):
    nx.square_clustering(G, backend="parallel")
```

so here the `n_jobs=4` is overwritten by the internal `with joblib.parallel_config(n_jobs=6)`(just for the nx-parallel algorithm). This is what the `_set_nx_config` decorator is doing.

dschult: Does `_set_nx_config` overwrite all the parameter settings of `parallel_config`? or only the values explcitily set in `nx_config`? For example:
```python
nx.config.backends.parallel.active = True
nx.config.backends.parallel.n_jobs = 6

with joblib.parallel_config(n_jobs=4, verbose=55):
    nx.square_clustering(G, backend="parallel")
```
Does this get run with `verbose=55`?

Schefflera-Arboricola: > Does this get run with verbose=55?

no, all of them get overwritten bcoz we pass all of the configs in `nx.config.backends.parallel` in the internal `parallel_config` run by the decorator(only when `active=True`). But, this is just for the nx-parallel functions, like here:

```py
with joblib.parallel_config(n_jobs=4, verbose=55):
    nx.square_clustering(G, backend="parallel") # n_jobs=6, verbose=0
    joblib.Parallel(joblib.delayed(sqrt)(i ** 2) for i in range(10)) # n_jobs = 4, verbose=55
```

dschult: Hmmm...  that seems confusing to me...

What's the advantage of making nx_config "active" over just having the user use `joblib.parallel_config`?
I guess they could specify different config parameters for the networkx parts and for any non-networkx code they are running... but that doesn't seem very persuasive. They have to work around that for the rest of their code. Why make nx-parallel different from a config perspective?

dschult: This does make sense...  Perhaps you can put this description (with an example) into the docs -- and explain that people can use joblib's config directly, or if they prefer the nx_config interface they can use that without affecting their other joblib code (because nx-parallel uses a context manager with the nx_config options).

dschult: Yes -- I see what you are saying now -- some users will be used to storing their backend configuration parameters using the nx_config system. They will find it natural to use the nx_config system to provide the config parameters to joblib through nx-parallel. Other users may be more comfortable using the joblib parallel_config system. By providing this option we support either method of configuration.
(written last night but I didn't press "comment" :)

nx_parallel/utils/decorators.py
```diff
@@ -0,0 +1,36 @@
...
+import networkx as nx
+from functools import wraps
+import os
+
+__all__ = [
+    "_configure_if_nx_active",
+]
+
+
+def _configure_if_nx_active():
```
Schefflera-Arboricola, line 12: We agreed on this name in the last meeting. (ref. [Notes](https://hackmd.io/@Schefflera-Arboricola/By6_8lTmR))

pyproject.toml
```diff
@@ -86,7 +86,7 @@ line-length = 88
 target-version = 'py310'
 
 [tool.ruff.lint.per-file-ignores]
-"__init__.py" = ['I', 'F403']
+"__init__.py" = ['I', 'F403', 'F401']
```
Schefflera-Arboricola, line 89: `'F401'` is added here to not get "module imported but unused." while running pre-commit on the `nx_parallel/__init__.py` because we just import `_configure_if_nx_active` in the file to include it in the public API but we don't use it.

nx_parallel/__init__.py
```diff
@@ -1,5 +1,6 @@
+from .utils.decorators import _configure_if_nx_active
```
Schefflera-Arboricola, line 1: As discussed in the last meeting, here I've added the `_configure_if_nx_active` in the nx_parallel's public API, mimicking how networkx adds the `_dispatchable` decorator in the networkx's public API by explicitly importing it in the `__init__.py` file

_nx_parallel/__init__.py
Schefflera-Arboricola: ref. https://github.com/networkx/nx-parallel/pull/75#discussion_r1728620575

Config.md
```diff
@@ -64,29 +64,67 @@ with nxp_config(n_jobs=7, verbose=0):
...
-- **Usage**: The main difference is how `backend_params` are passed. In `nx.config` you need to pass them as a dictionary, whereas in `joblib.parallel_config` you can just pass them along with the other configurations.
-- **Default Settings**: By default, `nx-parallel` looks for configs set by `joblib.parallel_config`, and NetworkX's config system needs to be enabled explicitly by making the `nx.config.backends.parallel.active` configuration `True`.
+- **Default Behavior**: By default, `nx-parallel` looks for configs in `joblib.parallel_config` unless `nx.config.backends.parallel.active` is set to `True`.
 
-### 3.2 When Should You Use Which System?
+### 3.3 When Should You Use Which System?
 
-- Use NetworkX's config system if you're working with multiple backends within NetworkX.
-- If you're using `nx-parallel` independently, you can choose either system based on your preference.
+- When you're working with multiple NetworkX backends, then to avoid having any conflicts with other backends' code, using networkx to configure nx-parallel would be well suited.
```
dschult, line 120: I think you have mentioned potential conflicts before. What are the potential conflicts?

Config.md
```diff
@@ -64,29 +64,67 @@ with nxp_config(n_jobs=7, verbose=0):
...
+- If using `nx-parallel` independently, either system can be used based on your preference.
 
-## 4. Important Notes
+## 4. Important Considerations
 
-- **What if you use both systems together?** If you try to use both config systems, NetworkX's configuration will take priority, and `joblib`'s settings will be ignored. To avoid this, make sure `networkx.config.backends.parallel.active` is set to `False` when using `joblib` to set configs, and `True` when using `networkx`.
-- **Why are there two config systems?** We're working towards syncing both systems in the future so that updating one will automatically update the other. An attempt to move towards a unified configuration system in nx-parallel was made by the PR#68, which introduced a custom context manager for `nx-parallel`.
+- **Avoiding Conflicts:** To prevent conflicts, ensure that `networkx.config.backends.parallel.active` is set to `False` when using `joblib`, and `True` when using NetworkX configurations.
```
dschult, line 125: Does "when using `joblib`" imply all use of joblib including in nx-parallel and outside nx-parallel? Or does this refer to using `joblib.parallel_config` to set the configuration for nx-parallel?

dschult (commented):
This is looking good -- I have a couple questions below. Probably tweaks in wording will clear it up.  Thanks...

Config.md
Schefflera-Arboricola: I have updated the `Config.md` with an example of this. Thanks!

Config.md
Schefflera-Arboricola: here, "when using joblib" means when the user is using `joblib.parallel_config` to configure nx-parallel's internal `joblib.Parallel()`.

Config.md
```diff
@@ -0,0 +1,138 @@
+# Configuring nx-parallel
+
+`nx-parallel` provides flexible parallel computing capabilities, allowing you to control settings like `backend`, `n_jobs`, `verbose`, and more. This can be done through two configuration systems: `joblib` and `NetworkX`. This guide explains how to configure `nx-parallel` using both systems.
+
+## 1. Using `joblib`
+
+`nx-parallel` relies on [`joblib.Parallel`](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html) for parallel computing. You can adjust its settings through the [`parallel_config`](https://joblib.readthedocs.io/en/latest/generated/joblib.parallel_config.html) class provided by `joblib`. For more details, check out the official [joblib documentation](https://joblib.readthedocs.io/en/latest/parallel.html).
```
dschult, line 7:
```suggestion
`nx-parallel` relies on [`joblib.Parallel`](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html) for parallel computing. You can adjust its settings through the [`joblib.parallel_config`](https://joblib.readthedocs.io/en/latest/generated/joblib.parallel_config.html) class provided by `joblib`. For more details, check out the official [joblib documentation](https://joblib.readthedocs.io/en/latest/parallel.html).
```

Config.md
```diff
@@ -0,0 +1,138 @@
+# Configuring nx-parallel
+
+`nx-parallel` provides flexible parallel computing capabilities, allowing you to control settings like `backend`, `n_jobs`, `verbose`, and more. This can be done through two configuration systems: `joblib` and `NetworkX`. This guide explains how to configure `nx-parallel` using both systems.
+
+## 1. Using `joblib`
```
dschult, line 5:
```suggestion
## 1. Using `joblib`'s config system
```

Config.md
```diff
@@ -0,0 +1,138 @@
...
+# Setting configs in a context
+with parallel_config(n_jobs=7, verbose=0):
+    nx.square_clustering(H)
+```
+
+Please refer the [official joblib's documentation](https://joblib.readthedocs.io/en/latest/generated/joblib.parallel_config.html) to better understand the config parameters.
+
+Note: Ensure that `nx.config.backends.parallel.active = False` when using `joblib` for configuration, as NetworkX configurations will override `joblib.parallel_config` settings if `active` is `True`.
+
+## 2. Using `networkx`
```
dschult, line 27:
```suggestion
## 2. Using `networkx`'s backend config system
```

Config.md
```diff
@@ -0,0 +1,138 @@
...
+```python
+import networkx as nx
+
+print(nx.config)
+```
+
+Output:
+
+```
+NetworkXConfig(backend_priority=[], backends=Config(parallel=ParallelConfig(active=False, backend='loky', n_jobs=None, verbose=0, temp_folder=None, max_nbytes='1M', mmap_mode='r', prefer=None, require=None, inner_max_num_threads=None, backend_params={})), cache_converted_graphs=True)
```
dschult, line 44: The output in the markdown is hard to read. Is there a better way to show this? Here's one attempt. But there is probably a better way -- don't worry about it too much. We can always tweak it later.
```suggestion
print(nx.config.backends.parallel)
```

Output:

```
ParallelConfig(active=False, backend='loky', n_jobs=None, verbose=0, temp_folder=None, max_nbytes='1M', mmap_mode='r', prefer=None, require=None, inner_max_num_threads=None, backend_params={})
```

Config.md
```diff
@@ -0,0 +1,138 @@
...
+nxp_config.verbose = 50
+
+nx.square_clustering(H)
+
+# Setting config in a context
+with nxp_config(n_jobs=7, verbose=0):
+    nx.square_clustering(H)
+```
+
+The configuration parameters are the same as `joblib.parallel_config`, so you can refer to the [official documentation](https://joblib.readthedocs.io/en/latest/generated/joblib.parallel_config.html) to better understand these config parameters.
```
dschult, line 67:
```suggestion
The configuration parameters are the same as `joblib.parallel_config`, so you can refer to the [official joblib documentation](https://joblib.readthedocs.io/en/latest/generated/joblib.parallel_config.html) to better understand these config parameters.
```

Config.md
```diff
@@ -0,0 +1,138 @@
...
+nx.config.backend_priority = ["another_nx_backend", "parallel"]
+nx.config.backends.another_nx_backend.config_1 = "xyz"
+joblib.parallel_config(n_jobs=7, verbose=50)
+
+nx.square_clustering(G)
+```
+
+In this example, if `another_nx_backend` also internally utilizes `joblib.Parallel` (without exposing it to the user) within its implementation of the `square_clustering` algorithm, then the `nx-parallel` configurations set by `joblib.parallel_config` will influence the internal `joblib.Parallel` used by `another_nx_backend`. To prevent unexpected behavior, it is advisable to configure these settings through the NetworkX configuration system.
+
+If you're using `nx-parallel` independently, either the NetworkX or `joblib` configuration system can be used, depending on your preference.
```
dschult, line 132: addressing the question of what "independently" means: one attempt
```suggestion
If your code using `nx-parallel` does not simultaneously use joblib (including inside another networkx backend) then either the `networkx.config` system or the `joblib.parallel_config` system can be used, depending on your preference.
```

Also, I think this paragraph should appear at the top of this section, as I think most users will be in this situation.

README.md
```diff
@@ -107,6 +107,13 @@ Note that for all functions inside `nx_code.py` that do not have an nx-parallel
 import networkx as nx
 import nx_parallel as nxp
 
+# enabling networkx's config for nx-parallel
```
dschult, line 110: I'm wonder if putting config issues into this section is a ds=istraction. They are trying to understand backend usage, not how to configure. Maybe a comment line pointing to the config docs here is sufficient. If you want to keep this here, there should at least be a word indicating that the config stuff is optional. 
```suggestion
# optional enabling networkx's config for nx-parallel
```

dschult (commented):
I took another look at the whole PR and the files rather than the diffs. So I've got a few more suggestions to think about. Adapt or reject them as you see fit.

README.md
Schefflera-Arboricola: I don't think it is optional. We don't get any errors when we run with the default configs. But, we are also not running the algorithms on multiple cpu cores or threads because `n_jobs=None` by default. So, to actually run an algorithm as a parallel algorithm they would have to set atleast the `n_jobs` config to something non 0 or non None. I have updated the README.md accordingly.

dschult: Wait -- the default value gives no parallel computation?  That doesn't seem right to me. Shouldn't the default be to use all the cpus we can get?

Schefflera-Arboricola: > Wait -- the default value gives no parallel computation? That doesn't seem right to me.

I understand your concern. However, this behavior is consistent with `joblib.parallel_config`, where parallel computation isn’t enabled by default (i.e. `n_jobs=None` by default). Since modifying the defaults within `joblib.parallel_config` is outside my control(bcoz that would mean changing joblib's codebase), changing the default `n_jobs` in NetworkX’s configuration from `None` to `-1` will introduce inconsistencies between the two configuration systems, and increase differences between the two config systems, leading to user confusion and potential pitfalls.

> Shouldn't the default be to use all the cpus we can get?

That was indeed the previous default, but I've since updated the documentation to reflect the current behavior. Reverting this change would necessitate another extensive documentation update across the entire package, which might not be feasible before the GSoC deadline.

Config.md
dschult: Is there any chance of trouble when the other backend doesn't use joblib?
For example, could the config using `joblib.parallel_config` interfere in any way with nx-cugraph or graphblas-algorithms?

I *think* the only way to have a "conflict" is if the other code is using joblib too. Right?

dschult: I think it doesn't matter what the codebases of the other backends are: the only config options nx-parallel uses are related to joblib. And the only way for another backend to be influenced by those choices is if they look at or use the config options in joblib. So only backends which use joblib will have a possible conflict over config values. 

In the interest of putting this into place, I am going to merge this PR. And I'll open another one to make further suggestions.

dschult (approved):
I approve of this PR.

I will make some suggested wording changes about when to use which config system once I figure out what they should be. But there is no point in delaying this PR for those changes. 

Thanks very much @Schefflera-Arboricola for putting this together and guiding it through!!

README.md
SH20RAJ: That sounds great

PR #78, opened 2024-08-20
Ignoring functions in `utils` for the `get_info` dict

Schefflera-Arboricola:
ref. https://github.com/networkx/nx-parallel/pull/75#discussion_r1723831104


dschult (approved):
This looks good to me! 

It means we only include `get_info` information for functions outside of the `utils` subpackage. If that kind of restriction becomes troublesome later we can implement a more precise method to exclude functions from the dict.  
 
Thanks!

PR #80, opened 2024-09-04
Chore/style cleanup utility functions

dPys:
## Summary

This PR focuses on stylistic improvements and cleanup of utility functions.

### Goal
Ensure better package maintainability, scalability, and ease of contribution for new developers.

### Code Style Adjustments
- Refactored variable names and formatting to improve readability.
- Improved indentation, spacing, and overall syntax compliance with PEP8.

### Utility Function Cleanup
- Introduced `@assign_algorithms` decorator to dynamically assign algorithms to `BackendInterface`, reducing manual updates and boilerplate.
- Simplified `BackendInterface` by replacing explicit algorithm declarations with automatic mapping from the `ALGORITHMS` list.
  - Even `ALGORITHMS` could be obviated by dynamically retrieving algorithm names directly from the modules using introspection (`__str__`, `dir()`, or `__dict__`), which might be too "auto-magical" for now but worth considering for a future PR.
- Improved `ParallelGraph`'s readability and consistency.
- Reduced documentation and maintenance overhead through a more scalable, dynamic approach to algorithm registration.
- Minor logic refactoring for efficiency.
- Fixed bug in chunking:
image

### Removal of Unused Imports/Code
- Eliminated unused imports, reducing code complexity and improving load time.

@dpys


_nx_parallel/script.sh
Schefflera-Arboricola, line 1: this script is run by the pre-commit hook that updates the `get_info` in the `_nx_parallel/__init__.py` file. So, that's why the style is failing. But, why have you removed the script?

dPys: I think it could have been failing before I modified it to have executable perms (I don't believeI removed it)?

dPys: Either way, just tried reverting that file to its original state of `main`

dPys: Okay @Schefflera-Arboricola -- I figured out what was happening. Just needed to update path check logic in `update_get_info.py` to prevent functions from the utils submodule from being added to the generated `__init__.py` file.

should be good to go!

dPys:
@Schefflera-Arboricola -- all type annotations have now been removed. If there are no other issues, then it might be helpful for us to get this merged sooner rather than later so that I can begin working on some of the other features that we discussed. WDYT?

nx_parallel/interface.py
```diff
@@ -1,125 +1,114 @@
...
+    "approximate_all_pairs_node_connectivity",
+    # Connectivity
+    "connectivity.all_pairs_node_connectivity",
+]
+
 
 class ParallelGraph:
-    """A wrapper class for networkx.Graph, networkx.DiGraph, networkx.MultiGraph,
-    and networkx.MultiDiGraph."""
+    """
```
dschult, line 48: I agree that we need the doc_string final `"""` to be on its own line (otherwise the doc_string doesn't end with an EOL char). But I also think we should not have the starting `"""` on its own line. That adds a blank line at the beginning of the doc_string. And sphinx looks at the first line to find text that is used when creating a link to that doc_string. I guess we aren't using sphinx for docs in this library. But it seems better to be uniform with doc_string spacing.

I'm suggesting the first line of test appear with the starting `"""` and the final `"""` be on a line with no text. But this is a nitpicky choice so I'm fine if this goes in as is.

nx_parallel/utils/chunk.py
```diff
@@ -2,65 +2,96 @@
...
-    G : NetworkX graph
-    iterator : str
-        Type of iterator. Valid values are 'node', 'edge', 'isolate'
-
-    Returns:
-    --------
-    iterable : Iterable
+    ----------
+    G : networkx.Graph
+        The NetworkX graph.
```
dschult, line 65: I would prefer not to put a specific class type here because it could be a DiGraph or MultiGraph, etc. How about `NetworkX graph`?

nx_parallel/utils/chunk.py
```diff
@@ -2,65 +2,96 @@
...
+    iterator : GraphIteratorType
+        Type of iterator. Valid values are 'NODE', 'EDGE', 'ISOLATE'.
```
dschult, line 67: The process of creating the enum GraphIteratorType makes this much harder to read -- involving looking across modules only to discover that they are just strings. Let's just use strings. Is there a reason for the extra complexity? If you have ideas of how this might grow to become worth the effort then can you explain that expectation?  The confusion between lowercase values and upper case attribute names seems like a C idiom, but could be confusing to some python users.

Also, while I believe the user could use strings as the inputs, that is not at all clear from the docs. If we stay with the enum approach then let's make these docs set up for string input.

Finally, should these be plural instead of singular? As in "iterator over nodes" or "iterator yields nodes". Not a blocker of course, but maybe we should think about it.

dschult (commented):
I don't feel strongly about any of these comments, but these are some suggestions for choices during the style cleanup.

dPys:
> I don't feel strongly about any of these comments, but these are some suggestions for choices during the style cleanup.

Thank you so much for the review! I've committed your formatting suggestions (though it appears now the linter test is failing?) and agree with all of your comments. It seems like the only thing that might warrant further discussion is the use of enums.

nx_parallel/utils/chunk.py
dPys: Using enums could be worth the trade-off for the type safety / clarity. They help prevent errors from typos in string literals, ensure consistency, and make the code more self-documenting and extendable for future iterations. That being said, I definitely see your point about keeping things simple and reducing overhead, especially if the set of iterator types isn't expected to grow. One potential compromise would be to allow both enums and strings, which could provide the best of both worlds—type safety when desired, and simplicity for users who prefer the string-based approach. 

Regarding the suggestion about plural vs. singular names (e.g., "nodes" instead of "node"), I think that's a great point. Using the plural form seems more intuitive since it better reflects what the iterator actually returns—multiple nodes, edges, or isolates. I've updated the iterator type names to be plural for consistency and clarity.

Like with the type annotations, I don't feel strongly about enum handling so am also happy to remove the enums for this PR altogether if it would expedite review. WDYT?

nx_parallel/interface.py
dPys: Absolutely! Fixed.

nx_parallel/utils/chunk.py
dPys: Sounds good, changed

nx_parallel/interface.py
```diff
@@ -1,125 +1,114 @@
...
             self.graph_object = nx.Graph()
-        elif isinstance(
-            graph_object, (nx.Graph, nx.DiGraph, nx.MultiGraph, nx.MultiDiGraph)
-        ):
+        elif isinstance(graph_object, NX_GTYPES):
             self.graph_object = graph_object
         else:
             self.graph_object = nx.Graph(graph_object)
 
+    @property
```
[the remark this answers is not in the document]
dPys: I had originally added the @property decorators here to make it easier to access `is_multigraph` and `is_directed` without needing parentheses, making the code feel simpler and more like you're just checking an attribute.

But indeed, this difference in accessing the `is_multigraph` and `is_directed` properties between `ParallelGraph` and `nx.Graph` introduces inconsistency. To maintain compatibility with the `nx.Graph` API, I've gone ahead and reverted this change, ensuring the same access pattern for both types of graphs.

nx_parallel/utils/chunk.py
```diff
@@ -1,66 +1,99 @@
...
     it = iter(iterable)
-    while True:
-        x = tuple(itertools.islice(it, n))
-        if not x:
-            return
-        yield x
+    for _ in range(n_chunks):
+        chunk_size = k + (1 if m > 0 else 0)
+        m -= 1
+        yield tuple(itertools.islice(it, chunk_size))
```
Schefflera-Arboricola, line 20: The current behaviour of this function is changed here. Was that intended by you or is this a bug?

```py
>>> data = list(range(11))
>>> list(new_chunks(data, 4))
[(0, 1, 2), (3, 4, 5), (6, 7, 8), (9, 10)]
>>> list(old_chunks(data, 4))
[(0, 1, 2, 3), (4, 5, 6, 7), (8, 9, 10)]
```

Schefflera-Arboricola (commented):
Thanks @dPys and @dschult!
I have added some high-level comments. LMKWYT. 
And thanks for removing type-annotations for now from this PR. We can consider adding them in future based on our need but for now we decided to not have them as they'll probably increase the barriers for the first-time-contributors.

_nx_parallel/script.sh
Schefflera-Arboricola: It showed `Empty file` on GitHub earlier. Also I'm not sure how checking `"nx_parallel/utils" in root` instead of checking `nx-parallel/nx_parallel/utils` at the end fixes this issue.

nx_parallel/utils/decorators.py
```diff
@@ -1,17 +1,19 @@
...
-import networkx as nx
 from functools import wraps
-import os
 
-__all__ = [
-    "_configure_if_nx_active",
-]
+import networkx as nx
+
+from joblib import parallel_config
```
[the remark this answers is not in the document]
dPys: Hmm, these changes were not made manually, so my guess was that it was either my VScode linter (set to `black`) and/or the pre-commit hooks

nx_parallel/utils/tests/test_chunk.py
```diff
@@ -0,0 +1,57 @@
+import os
```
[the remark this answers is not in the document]
dPys: Yes, this repo needs a lot more unit testing, so will need to find a way to prioritize that in the roadmap

nx_parallel/utils/chunk.py
```diff
@@ -1,66 +1,99 @@
...
+    If running under pytest, returns 2 jobs. If using a parallel backend,
+    returns the number of jobs configured for the backend. Otherwise, returns
+    the number of CPUs adjusted for negative `n_jobs` values.
+    """
     if "PYTEST_CURRENT_TEST" in os.environ:
         return 2
-    else:
+
+    if n_jobs == 0:
+        raise ValueError("n_jobs == 0 in Parallel has no meaning")
```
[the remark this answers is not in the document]
dPys: Moving back to end!

nx_parallel/utils/decorators.py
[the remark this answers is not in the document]
Schefflera-Arboricola: We should probably update the pre-commit hooks then :)

nx_parallel/utils/chunk.py
Schefflera-Arboricola: In today's discussion, @dPys and I figured that the docstrings indeed says that the `new_chunk` is the way we chunk, so this is more of a bug getting fixed. 

@dschult do you have any thoughts on what should be the default chunking - making chunks of size `n` or `n` number of chunk? (We do pass `n_jobs` as `n` in the chunking function, so it does make sense to have `n` chunks instead of chunks of size `n`). Also it would be great if you could give a final review as this PR is almost ready to be merged! Thank you :)

dPys: Screenshot 2024-09-20 at 2 48 31 PM
The above plot is for the new chunking -- runtime does not change as a function of graph size in the case of betweenness centrality. 

I had trouble even getting the old (i.e. current) chunking to work consistently across all configurations of chunks / n_cores, otherwise I'd post its benchmarks too. Can try to see if I can figure out why...

dschult: I think n-chunks is what we should be providing (to match n_jobs). I think it doesn't really matter what size the chunks are -- bigger will just take longer.  It is the number of chunks that is restricted. So that's a go from my perspective.

Nice plots of time vs number of nodes!  (note that graph "size" is technically the number of edges. graph "order" is the number of nodes, but "number of nodes" seems to be the term most people use.)  
It looks good up to N=250. But that's not actually that big. Still, it shows the trends that we want. And it is clear that we need N>250 for parallel to pay off in execution time in this situation. The memory plot seems to suggest that the number of cores doesn't affect memory. 

Maybe these plots should go in a main comment so it doesn't get "hidden" when the discussion is "resolved". :)

dschult (commented):
This is indeed getting close. :) 
I think the main thing is to make sure you are using the pre-commit linting for this package.  That way we aren't doing/undoing linting choices. That may not match the standard tools you use in your IDE. I hope it is not tricky to get working.  But it does seem that many of the changes here are lint related. 

The chunking tools look good to me.  I have a number of comments below (and github says I have one more than it is showing me, so I might have to fix/remove and old one from this review after submitting :).

.gitignore
```diff
@@ -135,3 +135,6 @@ html/
 
 # get_info update script
 temp__init__.py
+.history/
+
+.DS_Store
```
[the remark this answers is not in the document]
dPys: So these are relatively standard files/directories to include in a .gitignore (e.g. .DS_Store is actually in the .gitignore of the networkx core repo: https://github.com/networkx/networkx/blob/main/.gitignore#L4-L6). Maybe the best thing to do is just rote copy the .gitignore of networkx core identically?

nx_parallel/interface.py
```diff
@@ -1,50 +1,64 @@
...
-    is_reachable,
-    tournament_is_strongly_connected,
-)
-from nx_parallel.algorithms.vitality import closeness_vitality
-from nx_parallel.algorithms.approximation.connectivity import (
-    approximate_all_pairs_node_connectivity,
-)
-from nx_parallel.algorithms.connectivity import connectivity
-from nx_parallel.algorithms.cluster import square_clustering
+from typing import Union
```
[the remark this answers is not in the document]
dPys: The current state of the repo explicitly checks: `isinstance(graph_object, (nx.Graph, nx.DiGraph, nx.MultiGraph, nx.MultiDiGraph))`. If what we're saying is that we do not need to check each of these graph types explicitly (i.e. including the way I have here modified it using `NX_GTYPES` (with `Union` or `|`), and that it will suffice to simply check `isinstance(G, nx.Graph)`, then this indeed makes things way simpler!

nx_parallel/interface.py
```diff
@@ -1,50 +1,64 @@
...
+    and networkx.MultiDiGraph.
+    """
 
     __networkx_backend__ = "parallel"
 
-    def __init__(self, graph_object=None):
+    def __init__(
+        self,
+        graph_object,
+    ):
```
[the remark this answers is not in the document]
dPys: Good catch. The default has been added back-- it looks like it got removed on the last commit.

Regarding the linting, it must have been IDE-related change, but if that were the case, then the pre-commit hook should have failed, but it passed:
```
(base) bash-3.2$ pre-commit run --all-files --show-diff-on-failure --color always
Update function info.....................................................Passed
blacken-docs.............................................................Passed
prettier.................................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
```

so it appears that the pre-commit script needs some refinement as it's not enforcing the linting consistently. I've manually changed this back to a one-liner for now. Seems like we might want to open a new issue / PR to address the pre-commit handling more strictly?

nx_parallel/utils/chunk.py
dPys: Sounds good, removed and reverted for now

dschult:
What is the program that is creating the `.history` files?  :)

nx_parallel/utils/decorators.py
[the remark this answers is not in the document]
dschult: These import changes are made by some linter. Let's make sure we understand where that's coming from. I think in NetworkX we switched away from black toward ruff.  Maybe that's the cause.
Can your IDE get its linter switched depending on which project you are working with?

nx_parallel/utils/chunk.py
```diff
@@ -1,66 +1,102 @@
...
+    list_of_iterator : list, optional
+        A precomputed list of items to iterate over. If None, it will
+        be generated based on the iterator type.
 
-    Returns:
-    --------
-    iterable : Iterable
+    Returns
+    -------
+    iterable
```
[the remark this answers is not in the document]
dPys: This was a typo on my part. After removing the type annotations-- I forgot to revert to the original specification in the docstring. Good catch!

nx_parallel/utils/chunk.py
```diff
@@ -1,66 +1,102 @@
...
+        chunk_size = k + (1 if m > 0 else 0)
+        m -= 1
+        yield tuple(itertools.islice(it, chunk_size))
 
 
 def get_n_jobs(n_jobs=None):
-    """Returns the positive value of `n_jobs`."""
+    """Returns the positive value of `n_jobs`
+
+
```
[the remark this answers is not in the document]
dPys: This appears to have been a remanent of committing the doc code change suggestion in the UI. Fixed!

nx_parallel/utils/chunk.py
```diff
@@ -1,66 +1,102 @@
 import itertools
 import os
+
```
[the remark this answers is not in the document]
dPys: The blank lines are added automatically by the code formatter to group imports into standard library and third-party packages for readability. This is a common convention to make it easier to distinguish dependencies. Happy to remove it if that is the preference (I've done so on the latest commit)

nx_parallel/tests/test_get_chunks.py
```diff
@@ -2,11 +2,12 @@
...
 import inspect
 import importlib
-import networkx as nx
-import nx_parallel as nxp
 import random
 import types
 import math
+import networkx as nx
+
+import nx_parallel as nxp
```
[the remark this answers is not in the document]
dPys: The ordering is based on grouping: standard library imports (os), third-party imports (pytest, networkx), and local imports (nx_parallel), which AFAIK in standard in Python? The linter for nx-parallel is `black`, which is what I've also set in my IDE,  and it like other common linters (e.g. iSort, flake8, ruff) enforce these grouping ordering for readability, not strict alphabetical order. I've gone ahead and removed the extra-spacings for now, and I suppose we can return to the nitty gritty styling rules on a separate PR dedicated to [addressing the pre-commit issue](https://github.com/networkx/nx-parallel/issues/81)?

dPys:
> I think we're getting bogged down here in linting land... :( Can we change the `.gitignore` file in a separate PR?
Yes, happy to revert the .gitignore to `main` :) We can address this on a separate PR as well once this PR gets over the finish line.

dPys:
> What is the program that is creating the `.history` files? :)

The Local History extension of VScode creates .history folders to store automatic backups of your files so you can revert to previous versions, similar to how PyCharm's Local History works.

nx_parallel/utils/chunk.py
```diff
@@ -2,65 +2,94 @@
...
+    if n_jobs == 0:
+        raise ValueError("n_jobs == 0 in Parallel has no meaning")
+
+    return int(n_jobs)
 
 
 def create_iterables(G, iterator, n_cores, list_of_iterator=None):
-    """Creates an iterable of function inputs for parallel computation
+    """Create an iterable of function inputs for parallel computation
```
dschult, line 56: More of these minor nit-picky issues:We don't seem to be consistent in what tense the verb leading off the doc_string should use. I think NetworkX has tried to settle on "Create" over "Creates". But there are other doc_strings in this PR (like `get_n_jobs`) which use "Returns".  Rabbit holes abound. :)

nx_parallel/.DS_Store
dschult, line 1: This `.DS_Store` made it into the PR.

dschult (commented):
This is basically ready to go as far as I'm concerned.
There are some comments below.

nx_parallel/utils/chunk.py
dPys: `create` (i.e. the imperative mood) is indeed the preferred convention 99.99% of the time. The only time I've ever seen the third-person singular form being used is in Java. 

That being said, "Creates" was actually not something that I changed on this PR -- that is what the docstring reads on the `main` branch currently! That being said, happy to change it ;)

nx_parallel/.DS_Store
dPys: Removed!

dPys:
@MridulS -- it seems like removing the ruff dependency with that [direct push to `main`](https://github.com/networkx/nx-parallel/commit/c73066fbfd1d123a95bc842e44b8221ebc0f0cdc#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L51) is causing the style / format check to unconditionally fail for all PR's? Was that something that you were planning to revert?

dschult (approved):
I approve these changes.
@Schefflera-Arboricola can you re-take a look?

Schefflera-Arboricola:
> @MridulS -- it seems like removing the ruff dependency with that [direct push to `main`](https://github.com/networkx/nx-parallel/commit/c73066fbfd1d123a95bc842e44b8221ebc0f0cdc#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L51) is causing the style / format check to unconditionally fail for all PR's? Was that something that you were planning to revert?

Can you try moving the update-get-info precommit hook at the end in `.pre-commit-config.yaml` file, because that hook internally runs a ruff command and @MridulS removed the ruff dependency. But, I think if we run the ruff hook before update-get-info hook then ruff would be installed in the pre-commit environment for the update-get-info hook to use. I think this should resolve the failing style test. Also, I did the same thing in the PR that @MridulS made but that PR is made from the main branch of nx-parallel so i think that's the reason the style test is passing for the `push` but not for the `pull_request`. (PR https://github.com/networkx/nx-parallel/pull/84)

Schefflera-Arboricola (approved):
Thank you very much @dPys and @dschult  :)

I'm merging this even though the style CI is failing. We can handle that in another PR. This is good to go I think. Thank you again @dPys for putting this together and being so patient :)

PR #86, opened 2024-09-27
WIP: Refactor Parallel Graph Algorithms to Use a Centralized Parallel Configuration with Flexible Iterators

dPys:
## Summary:

This PR introduces a new **`execute_parallel`** function that obviates the need for separate joblib calls for each algorithm, obviates the old `create_iterables` logic, exposes the joblib parallel configuration arguments to the nx_parallel backend config.

## Key Updates:

1. **Add `execute_parallel`**:
   - Centralizes logic for parallel execution across data chunks.
   - Uses `iterator_func` to customize how data (nodes, edges, etc.) is iterated over.
   - Supports both default and custom chunking via `get_chunks`.

2. **Remove `create_iterables`**:
   - Simplified by moving to `iterator_func` within `execute_parallel`.

3. **Thread-safe Joblib config**:
   - Added a `parallel_config` context manager that uses thread-local storage to manage Joblib settings (like backend/verbose) without interference during concurrent runs.

4. **Refactor `betweenness_centrality` & `edge_betweenness_centrality`** to use `execute_parallel` as a POC.

## Why This Matters:

- **Parsimony**: Reduces code complexity by consolidating parallel execution logic into one function, making the codebase cleaner and easier to maintain.
- **Flexibility**: `execute_parallel` and `iterator_func` make this adaptable for any graph algs that need parallelism.
- **Efficiency**: Centralized parallel logic reduces duplication.
- **Maintainability**: Easier to extend and manage without `create_iterables`.

## TODO:
- Add unit test for `execute_parallel` once the interface gets the green-light
- Perform regression test to ensure equivalent functionality
- Finish refactoring the other algorithms to incorporate `execute_parallel` in place of existing separate joblib calls

### DIFF
--- a/nx_parallel/algorithms/centrality/betweenness.py
+++ b/nx_parallel/algorithms/centrality/betweenness.py
@@ -1,4 +1,3 @@
-from joblib import Parallel, delayed
 from networkx.algorithms.centrality.betweenness import (
     _accumulate_basic,
     _accumulate_endpoints,
@@ -46,23 +45,28 @@ def betweenness_centrality(
     else:
         nodes = seed.sample(list(G.nodes), k)
 
-    n_jobs = nxp.get_n_jobs()
+    def process_func(G, chunk, weight, endpoints):
+        return _betweenness_centrality_node_subset(
+            G, chunk, weight=weight, endpoints=endpoints
+        )
 
-    if get_chunks == "chunks":
-        node_chunks = nxp.create_iterables(G, "node", n_jobs, nodes)
-    else:
-        node_chunks = get_chunks(nodes)
+    def iterator_func(G):
+        return G.nodes
 
-    bt_cs = Parallel()(
-        delayed(_betweenness_centrality_node_subset)(G, chunk, weight, endpoints)
-        for chunk in node_chunks
+    results = nxp.utils.chunks.execute_parallel(
+        G,
+        process_func=process_func,
+        iterator_func=iterator_func,
+        get_chunks=get_chunks,
+        weight=weight,
+        endpoints=endpoints,
     )
 
-    # Reducing partial solution
-    bt_c = bt_cs[0]
-    for bt in bt_cs[1:]:
-        for n in bt:
-            bt_c[n] += bt[n]
+    # Aggregation of partial results
+    bt_c = {}
+    for bt in results:
+        for n, value in bt.items():
+            bt_c[n] = bt_c.get(n, 0.0) + value
 
     betweenness = _rescale(
         bt_c,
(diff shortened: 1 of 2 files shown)


nx_parallel/algorithms/centrality/betweenness.py
```diff
@@ -41,28 +40,31 @@ def betweenness_centrality(
...
-    bt_cs = Parallel()(
-        delayed(_betweenness_centrality_node_subset)(G, chunk, weight, endpoints)
-        for chunk in node_chunks
+    def iterator_func(G):
+        if k is None:
+            return G.nodes
+        else:
+            return seed.sample(list(G.nodes), k)
+
+    bt_cs = nxp.utils.chunks.execute_parallel(
```
Schefflera-Arboricola, line 54:
```suggestion
    bt_cs = nxp.utils.chunk.execute_parallel(
```

nx_parallel/utils/chunk.py
```diff
@@ -52,44 +77,56 @@ def get_n_jobs(n_jobs=None):
...
+        data_chunks = get_chunks(data)
+    else:
+        raise ValueError(
+            "get_chunks must be 'chunks' or a callable that returns an iterable of chunks."
+        )
+
+    # read parallel_kwargs from thread-local storage
+    parallel_kwargs = getattr(_joblib_config, "parallel_kwargs", {})
+
+    return Parallel(n_jobs=n_jobs, **(parallel_kwargs or {}))(
```
Schefflera-Arboricola, line 130: already taken care by the decorator
```suggestion
    return Parallel()(
```

dschult:
These changes to the centralized processing of chunks seem good. I like the general `iterator_func` capability, and it still allows inclusion of common cases like "nodes".

So +1 on the the interface from me, though I'd like to hear from @Schefflera-Arboricola too. :)

Thanks for this!

nx_parallel/algorithms/centrality/betweenness.py
```diff
@@ -41,28 +41,32 @@ def betweenness_centrality(
     if hasattr(G, "graph_object"):
         G = G.graph_object
```
Schefflera-Arboricola, line 42: remove this

dPys: So this is actually present on `main`, and if we remove this here, the tests fails

dschult (approved):
This looks good to me!
Sorry for the delay in review.

dPys:
No problem! Looks like we still need @Schefflera-Arboricola to also approve before this can get merged.

PR #87, opened 2024-10-11
Bug: getting rid of the ParallelConfig class

Schefflera-Arboricola:
closes https://github.com/networkx/networkx/issues/7671

need to investigate more. This PR will only fix the warning showing on the nx website but if some backend is setting their config using the networkx's `Config` class then they will get this warning on their local machine. also the user won't be able to set configurations for that backend. And we should probably have a better way of testing if the entry_points are working fine for all the backend(in networkx not in nx-parallel).

Thank you :)


jarrodmillman (approved):
Thanks!!

dschult:
I think this won't be needed because the trouble has been fixed on the NetworkX side. 
Thanks for the quick response @Schefflera-Arboricola and workaround. :)

Schefflera-Arboricola:
fix - https://github.com/networkx/networkx/pull/7672

PR #89, opened 2024-10-15
TST: Added tests checking entry_points discovery and config initialisation

Schefflera-Arboricola:
and updated networkx version to `3.4.2`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,7 @@ jobs:
         run: |
           python -m pip install git+https://github.com/networkx/networkx.git@main
           python -m pip install .
+          python -m pip install ".[test]"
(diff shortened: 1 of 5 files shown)


dschult (approved):
I now understand why the order(1) needs to be first -- otherwise the nx.config values might have been changed by the other tests (there are side effects of the tests). 

And if I understand corretly the `order(2)` needs to be run before the others so we can test that the default number of `n_jobs` is 1.  

Should we be managing the nx.configs around all the tests? 
Maybe with an automated setup for each test that resets nx.config to its default values?
That might be simpler code-wise than relying on test order -- but it also might slow down the tests.

Anyway, I approve this PR as is and we can switch to a universal setup of the nx.config in another PR if we want.

Thanks!!

Issue #82, opened 2024-09-23
Adding Embarrassingly Parallel Algorithms in NetworkX to nx-parallel


Schefflera-Arboricola:
## Embarrassingly parallel algorithms

Embarrassingly parallel algorithms are those that can be easily and efficiently divided into smaller, independent tasks that can be executed concurrently across multiple processing units, such as CPUs, GPUs, or nodes, without requiring significant communication or synchronisation between them. This type of parallelism is often referred to as “embarrassingly parallel” because it’s so straightforward and easy to implement. Refer [this](https://joblib.readthedocs.io/en/latest/parallel.html) to see how to implement embarrassingly parallel algorithms using joblib.

Characteristics of Embarrassingly Parallel Problems(Algorithms):

- Independent tasks: Each task can be executed independently, without relying on the output of other tasks.
- No communication: Tasks do not require frequent or significant communication with each other.
- No data dependencies: Tasks do not rely on shared data or intermediate results from other tasks.
- Easy to divide: The problem can be easily split into smaller, parallelizable tasks.

### Step 1 : Find a NetworkX algorithm that is embarrassingly parallel

- You can find such an algorithm from the NetworkX's references - https://networkx.org/documentation/latest/reference/index.html
- Make sure there is a `_dispatchable` decorator(in the networkx's codebase) on the algorithm you have selected to parallelise.

### Step 2 : Complete the checklist in Contributor's guide

This [checklist](https://github.com/networkx/nx-parallel/blob/main/CONTRIBUTING.md#general-guidelines-on-adding-a-new-algorithm) in Contributor's guide for adding a new algorithm to the package.

## Some references:

It would be better to look at [currently existing algorithms](https://github.com/networkx/nx-parallel/tree/main/nx_parallel/algorithms) than previous PRs that add parallel algorithms(like https://github.com/networkx/nx-parallel/pull/60 , https://github.com/networkx/nx-parallel/pull/37 , https://github.com/networkx/nx-parallel/pull/45 and many more) because we have changed a few things. 

Feel free to ask questions if you get stuck anywhere or if anything is not clear or if you want something to be explained in a bit more detail here.

Thank you :)

RohitP2005:
Hey , i would like to work on this issue

Schefflera-Arboricola:
Hi @RohitP2005 , let us know if you have any questions :)

Also, writing parallel versions of functions that return a non-generator(i.e. functions that `return` and not `yield`) would be easier to begin with.

[...]

PR #91, opened 2024-12-18
added traversal algorithm to nx_parallel

RohitP2005:
I added traversal algorithm with dfs and bfs which is ```_dispatchable``` . Kindly suggest further corrections


RohitP2005:
@stefanv can u help with what to be done next

dschult (commented):
This is a lot of new functions!  
But I can't tell if it is doing anything that networkx is not already doing.
For example the `breadth_first_search.py` functions look copied straight fro networkx. That code is not parallel. I think the idea is to split up the "embarrasing parallelizable" loop and do the iterations of that loop on different processors. 

Maybe just focus on one function and get that working... Then you can do the others.
To get an idea of how this works look at the other functions already in nx-parallel. For example, `betweenness_centrality`.

RohitP2005:
> This is a lot of new functions! But I can't tell if it is doing anything that networkx is not already doing. For example the `breadth_first_search.py` functions look copied straight fro networkx. That code is not parallel. I think the idea is to split up the "embarrasing parallelizable" loop and do the iterations of that loop on different processors.
> 
> Maybe just focus on one function and get that working... Then you can do the others. To get an idea of how this works look at the other functions already in nx-parallel. For example, `betweenness_centrality`.

Sorry about that , i will look into it asap , thanks for the help

RohitP2005:
@dschult
Thanks so much for your detailed feedback! You're absolutely right—it’s better to focus on implementing and testing one function thoroughly before moving on to others.

I’ve removed the DFS algorithm for now and added a BFS algorithm that is parallelized for improved performance.

Here’s an example of how to use the new parallel_bfs function:
```
# Create a sample graph
G = nx.erdos_renyi_graph(100, 0.05)

# Perform parallel BFS
bfs_result = parallel_bfs(G, source=0)
print("BFS Traversal Order:", bfs_result)

```

Schefflera-Arboricola (commented):
Thanks @RohitP2005 for the PR! Could you please follow this checklist for adding a new algorithm to the project - https://github.com/networkx/nx-parallel/blob/main/CONTRIBUTING.md#general-guidelines-on-adding-a-new-algorithm and you can also look at the existing algorithms in the repository to get a better idea! If any of these points in the checklist are not clear then please let us know. Thanks!

RohitP2005:
@Schefflera-Arboricola Yeah sure , i will look into it.
I will make sure the new algorithm added meets your expectations

nx_parallel/algorithms/traversal/breadth_first_search.py
```diff
@@ -0,0 +1,96 @@
+from joblib import Parallel, delayed
+from networkx.utils import py_random_state
+import nx_parallel as nxp
+
+__all__ = ["parallel_bfs"]
+
+
+@nxp._configure_if_nx_active()
+@py_random_state(3)
+def parallel_bfs(G, source=None, get_chunks="chunks", n_jobs=None):
```
Schefflera-Arboricola, line 10: n_jobs is included in the configurations. so no need to include it as a parameter here. (see - https://github.com/networkx/nx-parallel/blob/main/Config.md)

Schefflera-Arboricola, line 10: the names of the functions in the main networkx project and in this nx-parallel project are usually kept same. (see - https://github.com/networkx/nx-parallel/blob/main/CONTRIBUTING.md#general-guidelines-on-adding-a-new-algorithm)

nx_parallel/algorithms/traversal/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .depth_first_search import *
```
Schefflera-Arboricola, line 1:
```suggestion
```

nx_parallel/algorithms/traversal/breadth_first_search.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    Perform a parallelized Breadth-First Search (BFS) on the graph.
+
+    Parameters
+    ----------
+    G : graph
+        A NetworkX graph.
+    source : node, optional
+        Starting node for the BFS traversal. If None, BFS is performed for all nodes.
```
Schefflera-Arboricola, line 19: you don't need to document the parameters that are already documented in the main networkx repo :)

Schefflera-Arboricola (commented):
Thanks for all the work @RohitP2005 !

I haven't looked at the core implementation in this PR too closely.. but since BFS is not an embarrassingly parallel algorithm, I don't think its parallel implementation could look this simple 😅. If you are new to parallel graph algorithms then I'd suggest you consider adding an embarrassingly parallel algorithm instead (see - https://github.com/networkx/nx-parallel/issues/82). BFS for starters might be not be great.

Thanks, and sorry for the delayed reviews :)

Issue #99, opened 2025-03-25
Ensure nx-parallel's consistency with NetworkX for empty graph handling


Schefflera-Arboricola:
## Current behaviour

```python
>>> G = nx.Graph()
>>> nx.betweenness_centrality(G)
{}
>>> nx.betweenness_centrality(G, backend="parallel")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/anaconda3/lib/python3.12/site-packages/networkx/utils/decorators.py", line 788, in func
    return argmap._lazy_compile(__wrapper)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<class 'networkx.utils.decorators.argmap'> compilation 5", line 4, in argmap_betweenness_centrality_1
  File "/opt/anaconda3/lib/python3.12/site-packages/networkx/utils/backends.py", line 1091, in __call__
    rv = self._convert_and_call(
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/lib/python3.12/site-packages/networkx/utils/backends.py", line 1947, in _convert_and_call
    return func(*converted_args, **converted_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/lib/python3.12/site-packages/nx_parallel/utils/decorators.py", line 30, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/lib/python3.12/site-packages/networkx/utils/decorators.py", line 788, in func
    return argmap._lazy_compile(__wrapper)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<class 'networkx.utils.decorators.argmap'> compilation 9", line 4, in argmap_betweenness_centrality_6
  File "/opt/anaconda3/lib/python3.12/site-packages/nx_parallel/algorithms/centrality/betweenness.py", line 62, in betweenness_centrality
    bt_c = bt_cs[0]
           ~~~~~^^^
IndexError: list index out of range
```

## Expected behaviour

```python
>>> G = nx.Graph()
>>> nx.betweenness_centrality(G)
{}
>>> nx.betweenness_centrality(G, backend="parallel")
[...]

PR #100, opened 2025-03-25
Handling empty graph edge-case for `betweenness_centrality` and `edge_betweenness_centrality`

akshitasure12:
This PR addresses issue #99 by adding a condition to return an empty dictionary `{}` for empty graphs in the following functions:

1. `betweenness_centrality`
2. `edge_betweenness_centrality`

I am planning to implement similar checks for other functions as well.
Let me know if I should tweak it any further!

### DIFF
--- a/nx_parallel/algorithms/centrality/betweenness.py
+++ b/nx_parallel/algorithms/centrality/betweenness.py
@@ -61,9 +61,6 @@ def betweenness_centrality(
         for chunk in node_chunks
     )
 
-    if not bt_cs:
-        return {}
-
     # Reducing partial solution
     bt_c = bt_cs[0]
     for bt in bt_cs[1:]:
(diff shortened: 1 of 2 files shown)


nx_parallel/algorithms/centrality/betweenness.py
```diff
@@ -58,6 +58,9 @@ def betweenness_centrality(
         for chunk in node_chunks
     )
 
+    if not bt_cs:
```
dschult, line 61: I think we should probably put this right after we update `G`.  For this one, that is line 43. We avoid extra work by checking `len(G) == 0` and returning `{}` in that case.  Thoughts?

akshitasure12:
I'm a bit unsure about how these checks are failing since the only change I made was moving the empty graph check. Could you help me understand if I might be missing something?

Issue #94, opened 2025-03-22
BUG: `get_all_functions()` is not returning `args` and `kwargs` of the functions in `algorithms`


akshitasure12:
Hi @Schefflera-Arboricola,
I added a new algorithm and wanted to try it out. While experimenting, I added two print statements to the `get_all_functions()` function in `test_get_chunks.py` to check the function's behavior. However, I encountered the following : 
```
platform darwin -- Python 3.12.4, pytest-7.4.4, pluggy-1.0.0
rootdir: /Users/a91986/Desktop/nx-parallel
plugins: anyio-4.2.0
collected 1 item                                                                                                                                                                                           

nx_parallel/tests/test_get_chunks.py all_pairs_all_shortest_paths
{'args': [], 'kwargs': 'args'}
all_pairs_bellman_ford_path
{'args': [], 'kwargs': 'args'}
all_pairs_bellman_ford_path_length
{'args': [], 'kwargs': 'args'}
all_pairs_dijkstra
{'args': [], 'kwargs': 'args'}
all_pairs_dijkstra_path
{'args': [], 'kwargs': 'args'}
all_pairs_dijkstra_path_length
{'args': [], 'kwargs': 'args'}
all_pairs_node_connectivity
{'args': [], 'kwargs': 'args'}
all_pairs_shortest_path
{'args': [], 'kwargs': 'args'}
all_pairs_shortest_path_length
{'args': [], 'kwargs': 'args'}
approximate_all_pairs_node_connectivity
{'args': [], 'kwargs': 'args'}
betweenness_centrality
{'args': [], 'kwargs': 'args'}
chunks
{'args': ['iterable', 'n_chunks'], 'kwargs': None}
closeness_vitality
{'args': [], 'kwargs': 'args'}
create_iterables
{'args': ['G', 'iterator', 'n_cores', 'list_of_iterator'], 'kwargs': None}
edge_betweenness_centrality
{'args': [], 'kwargs': 'args'}
get_n_jobs
{'args': ['n_jobs'], 'kwargs': None}
is_reachable
{'args': [], 'kwargs': 'args'}
johnson
{'args': [], 'kwargs': 'args'}
local_efficiency
{'args': [], 'kwargs': 'args'}
node_redundancy
{'args': [], 'kwargs': 'args'}
number_of_isolates
{'args': [], 'kwargs': 'args'}
square_clustering
{'args': [], 'kwargs': 'args'}
tournament_is_strongly_connected
[...]

PR #104, opened 2025-03-30
Refactor get_all_functions to use inspect.signature instead of inspect.getfullargspe

akshitasure12:
This PR addresses issue #94

Changes made : 
1. Replaced `inspect.getfullargspec()` with `inspect.signature()` for reporting the signature of the original function and not that of wrapped.

2. Updated the assertion logic in the `test_get_chunks` for `chk_dict_vals` to iterate over graph edges instead of nodes since `chk_dict_vals` consists of functions that deals with edges between the nodes and not the nodes themselves.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,11 +121,12 @@ The default chunking in nx-parallel is done by slicing the list of nodes (or edg
 - The algorithm that you are considering to add to nx-parallel should be in the main networkx repository and it should have the `_dispatchable` decorator. If not, you can consider adding a sequential implementation in networkx first.
 - check-list for adding a new function:
   - [ ] Add the parallel implementation(make sure API doesn't break), the file structure should be the same as that in networkx.
-  - [ ] add the function to the `BackendInterface` class in [interface.py](./nx_parallel/interface.py) (take care of the `name` parameter in `_dispatchable` (ref. [docs](https://networkx.org/documentation/latest/reference/backends.html)))
+  - [ ] Include the `get_chunks` additional parameter. Currently, all algorithms in nx-parallel offer the user to pass their own custom chunks. Unless it is impossible to chunk, please do include this additional parameter.
+  - [ ] add the function to the `ALGORITHMS` list in [interface.py](./nx_parallel/interface.py). Take care of the `name` parameter in `_dispatchable` for the algorithms with same name but different implementations. The `name` parameter is used distinguish such algorithms in a single namespace. (ref. [docs](https://networkx.org/documentation/latest/reference/backends.html)))
   - [ ] update the `__init__.py` files accordingly
   - [ ] docstring following the above format
-  - [ ] run the [timing script](./timing/timing_individual_function.py) to get the performance heatmap
-  - [ ] add additional test(if any)
-  - [ ] add benchmark(s) for the new function(ref. the README in benchmarks folder for more details)
+  - [ ] add additional test, if needed. The smoke tests for the additional parameter `get_chunks` are done [here](https://github.com/networkx/nx-parallel/blob/main/nx_parallel/tests/test_get_chunks.py) together for all the algorithms.
+  - [ ] run the [timing script](./timing/timing_individual_function.py) to get the performance heatmap (ref. [Issue#51](https://github.com/networkx/nx-parallel/issues/51))
+  - [ ] add benchmark(s) for the new function(ref. the README in `benchmarks` folder for more details)
 
 Happy contributing! 🎉
--- a/nx_parallel/tests/test_get_chunks.py
+++ b/nx_parallel/tests/test_get_chunks.py
@@ -22,7 +22,21 @@ def get_all_functions(package_name="nx_parallel"):
 
     for name, obj in inspect.getmembers(package, inspect.isfunction):
         if not name.startswith("_"):
-            args, kwargs = inspect.getfullargspec(obj)[:2]
+            signature = inspect.signature(obj)
+            args = [
+                param.name
+                for param in signature.parameters.values()
+                if param.kind
+                in (
+                    inspect.Parameter.POSITIONAL_ONLY,
+                    inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                )
+            ]
+            kwargs = [
+                param.name
+                for param in signature.parameters.values()
+                if param.kind == inspect.Parameter.KEYWORD_ONLY
+            ]
             functions[name] = {"args": args, "kwargs": kwargs}
 
     return functions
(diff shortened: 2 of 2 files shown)


nx_parallel/tests/test_get_chunks.py
```diff
@@ -22,7 +22,21 @@ def get_all_functions(package_name="nx_parallel"):
...
+                in (
+                    inspect.Parameter.POSITIONAL_ONLY,
+                    inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                )
+            ]
+            kwargs = [
+                param.name
+                for param in signature.parameters.values()
+                if param.kind == inspect.Parameter.KEYWORD_ONLY
+            ]
```
Schefflera-Arboricola, line 39: I checked the tests are passing(using `pytest nx_parallel`). The CI test failure is unrelated!

Maybe we can simplify this code by only returning kwargs instead of both args and kwargs as `get_chunks` is always a kwarg. Does that sound like a good idea to you?

nx_parallel/tests/test_get_chunks.py
```diff
@@ -76,14 +90,18 @@ def random_chunking(nodes):
                 if isinstance(c1, types.GeneratorType):
                     c1, c2 = dict(c1), dict(c2)
                     if func in chk_dict_vals:
-                        for i in range(len(G.nodes)):
-                            assert math.isclose(c1[i], c2[i], abs_tol=1e-16)
+                        for edge in G.edges:
+                            assert math.isclose(
+                                c1.get(edge, 0), c2.get(edge, 0), abs_tol=1e-16
+                            )
```
Schefflera-Arboricola, line 96: Good catch! But this will now work with `edge_betweenness_centrality` and not `betweenness_centrality` because there the keys are nodes. But, this seems like it is working because when you apply `get` on edges it always returns `0` for both c1 and c2. But, that shouldn't happen. So maybe we should use `c1.keys` instead of G.nodes(or G.edges)

CONTRIBUTING.md
```diff
@@ -121,11 +121,12 @@ The default chunking in nx-parallel is done by slicing the list of nodes (or edg
 - The algorithm that you are considering to add to nx-parallel should be in the main networkx repository and it should have the `_dispatchable` decorator. If not, you can consider adding a sequential implementation in networkx first.
 - check-list for adding a new function:
   - [ ] Add the parallel implementation(make sure API doesn't break), the file structure should be the same as that in networkx.
-  - [ ] add the function to the `BackendInterface` class in [interface.py](./nx_parallel/interface.py) (take care of the `name` parameter in `_dispatchable` (ref. [docs](https://networkx.org/documentation/latest/reference/backends.html)))
+  - [ ] Include the `get_chunks` additional parameter. Currently, all algorithms in nx-parallel offer the user to pass their own custom chunks. Unless it is impossible to chunk, please do include this additional parameter.
```
Schefflera-Arboricola, line 124: Could you consider rebasing this branch -- so that these changes don't appear in the diff? Thanks!

akshitasure12:
I agree with you @Schefflera-Arboricola! I've made the changes you suggested.
Let me know if anything else needs to be addressed!

nx_parallel/tests/test_get_chunks.py
```diff
@@ -18,22 +18,26 @@ def get_all_functions(package_name="nx_parallel"):
...
-            args, kwargs = inspect.getfullargspec(obj)[:2]
-            functions[name] = {"args": args, "kwargs": kwargs}
-
-    return functions
+            signature = inspect.signature(obj)
+            kwargs = [
+                param.name
+                for param in signature.parameters.values()
+                if param.default is not inspect.Parameter.empty
+            ]
```
Schefflera-Arboricola, line 30: this is collecting all the default values of all the kwargs-- not the kwargs. Is that what you intended? The `get_functions_with_get_chunks` function calls this function-- and it assumes that `get_all_functions` is providing all the functions keyed to their kwargs. Could you please consider updating the docstrings of these functions to make it more clear to review?


We should also consider adding these helper functions to test/utils or utils maybe and then testing these as well

nx_parallel/tests/test_get_chunks.py
```diff
@@ -76,14 +80,18 @@ def random_chunking(nodes):
                 if isinstance(c1, types.GeneratorType):
                     c1, c2 = dict(c1), dict(c2)
                     if func in chk_dict_vals:
-                        for i in range(len(G.nodes)):
-                            assert math.isclose(c1[i], c2[i], abs_tol=1e-16)
+                        for key in c1.keys():
+                            assert math.isclose(
+                                c1.get(key, 0), c2.get(key, 0), abs_tol=1e-16
```
Schefflera-Arboricola, line 85: I don't think we should use `.get` here because the keys should be the same in both c1 and c2 and if they are not-- then an error should be raised instead of using 0. LMKWYT.

nx_parallel/tests/test_get_chunks.py
```diff
@@ -76,14 +80,18 @@ def random_chunking(nodes):
...
+                            )
                     else:
                         assert c1 == c2
                 else:
                     if func in chk_dict_vals:
-                        for i in range(len(G.nodes)):
-                            assert math.isclose(c1[i], c2[i], abs_tol=1e-16)
+                        for key in c1.keys():
+                            assert math.isclose(
+                                c1.get(key, 0), c2.get(key, 0), abs_tol=1e-16
```
Schefflera-Arboricola, line 93: same here.

Schefflera-Arboricola (commented):
Thanks @akshitasure12 -- Apologies for the delayed response! Please LMK if you still have the bandwidth to address the following comments. Thanks!

akshitasure12:
> Thanks @akshitasure12 -- Apologies for the delayed response! Please LMK if you still have the bandwidth to address the following comments. Thanks!

Sure! I’ll be free in about a week and will get back to it.

nx_parallel/tests/test_get_chunks.py
akshitasure12: That was unintentional! I’ll get these changes done asap!

CONTRIBUTING.md
Schefflera-Arboricola: https://youtu.be/DkWDHzmMvyg?si=bzbbgKiZDUppp1uC

nx_parallel/tests/test_get_chunks.py
```diff
@@ -76,14 +80,20 @@ def random_chunking(nodes):
                 if isinstance(c1, types.GeneratorType):
                     c1, c2 = dict(c1), dict(c2)
                     if func in chk_dict_vals:
-                        for i in range(len(G.nodes)):
-                            assert math.isclose(c1[i], c2[i], abs_tol=1e-16)
+                        for key in c1:
+                            if not math.isclose(c1[key], c2[key], abs_tol=1e-16):
+                                raise ValueError(
+                                    f"Values for key '{key}' differ: {c1[key]} != {c2[key]}"
+                                )
```
Schefflera-Arboricola, line 87: no need for explicitly raising error like this here, assert already handles this

nx_parallel/tests/test_get_chunks.py
```diff
@@ -10,30 +10,34 @@
 import nx_parallel as nxp
 
 
-def get_all_functions(package_name="nx_parallel"):
-    """Returns a dict keyed by function names to its arguments.
+def get_all_functions_kwargs(package_name="nx_parallel"):
+    """Returns a dict keyed by function names to its positional-or-keyword arguments.
```
Schefflera-Arboricola, line 14: why have you explicitly mentioned "positional-or-keyword"? -- it returns all the arguments. also in the following code could you please update the variable names i.e. update `kwarg_`?

akshitasure12: All the functions only have positional or keyword arguments, so there is no need to mention that explicitly. I'll make the changes. Thanks!

Schefflera-Arboricola (approved):
Thanks @akshitasure12! 
CI failures unrelated.

dschult (approved):
This looks like a good simplification and refactor.
And, of course, it now avoids the problems with `getfullargspec`.

I approve this PR

Issue #105, opened 2025-04-07
Fix unknown `pytest.mark.order` warnings


Schefflera-Arboricola:
Get rid of these warnnings:

```
(base) nx-parallel $ pytest nx_parallel
================================================================= test session starts =================================================================
platform linux -- Python 3.12.1, pytest-8.3.5, pluggy-1.5.0
rootdir: /workspaces/nx-parallel
configfile: pyproject.toml
plugins: anyio-4.9.0
collected 8 items                                                                                                                                     

nx_parallel/algorithms/centrality/tests/test_betweenness_centrality.py .                                                                        [ 12%]
nx_parallel/tests/test_entry_points.py ...                                                                                                      [ 50%]
nx_parallel/tests/test_get_chunks.py .                                                                                                          [ 62%]
nx_parallel/utils/tests/test_chunk.py ...                                                                                                       [100%]

================================================================== warnings summary ===================================================================
nx_parallel/tests/test_entry_points.py:19
  /nx-parallel/nx_parallel/tests/test_entry_points.py:19: PytestUnknownMarkWarning: Unknown pytest.mark.order - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.order(1)

nx_parallel/utils/tests/test_chunk.py:7
  /nx-parallel/nx_parallel/utils/tests/test_chunk.py:7: PytestUnknownMarkWarning: Unknown pytest.mark.order - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.order(2)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
====================================================== 8 passed, 2 warnings in 205.58s (0:03:25) ======================================================

```

pakagronglb:
Hello. Would you mind if I take on this issue? I have it all setup on my local machine and will pr to you once its all done.

Schefflera-Arboricola:
Sorry-- I didn't install the `pytest-order` plug-in

`plugins: anyio-4.9.0, order-1.3.0`

also I did get this other warning

```
================================================================== warnings summary ===================================================================
nx_parallel/tests/test_get_chunks.py::test_get_chunks
nx_parallel/tests/test_get_chunks.py::test_get_chunks
nx_parallel/tests/test_get_chunks.py::test_get_chunks
[...]

PR #107, opened 2025-04-09
Added a custom marker to avoid `Unknown pytest.mark.order` warning

akshitasure12:
Addressing issue #105

This PR adds `pytest-order` to the developer dependencies in `pyproject.toml` so that`@pytest.mark.order` is recognized during test runs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,13 +49,17 @@ parallel = "_nx_parallel:get_info"
 developer = [
     'pre-commit==3.8.0',
     'ruff==0.6.7',
-    'pytest-order',
 ]
 test = [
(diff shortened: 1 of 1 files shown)


akshitasure12:
Apologies, I misunderstood the warning! In the latest commit, I added a custom marker as suggested [here](https://docs.pytest.org/en/stable/how-to/mark.html#:~:text=You%20can%20register%20custom%20marks%20in%20your%20pytest.ini%20file%20like%20this%3A). This fixes the warnings.

Schefflera-Arboricola:
Thanks @akshitasure12 ! 
But this isn't required; we are already using pytest-order.

PR #112, opened 2025-05-15
correctly pass n_jobs as n_chunks to chunks()

### LINKED ISSUE #110: Incorrect usage of `num_in_chunk` as number of chunks in `chunks()`
In `is_reachable()` under `tournament.py`,  while creating chunks of nodes from a graph G, we compute `num_in_chunk`.

```
if get_chunks == "chunks":
        num_in_chunk = max(len(G) // n_jobs, 1)
        node_chunks = nxp.chunks(G, num_in_chunk)
```

This chunk size is passed as `n_chunks` to the chunks function. 

```
def chunks(iterable, n_chunks):
    """Yield exactly `n_chunks` chunks from `iterable`, balancing the chunk sizes."""
    iterable = list(iterable)
```

## Expected Behavior : 

```
if get_chunks == "chunks":
    node_chunks = nxp.chunks(G, n_jobs)
else:
    node_chunks = get_chunks(G)
```
This chunking logic pattern is used in several other functions as well. Let me know what you think!

akshitasure12:
Fixes #110 

Between the two approaches:
1. Passing `n_jobs` to `chunks()`.
2. Modifying `chunks()` to accept `num_in_chunk` instead of `n_chunks`.
I chose Approach 1 — passing `n_jobs` as `n_chunks` — because this allows `chunks()` to internally compute balanced chunk sizes, handling cases where the total number of nodes cannot be evenly divided, ensuring a more flexible distribution of work.

### DIFF
--- a/nx_parallel/algorithms/approximation/connectivity.py
+++ b/nx_parallel/algorithms/approximation/connectivity.py
@@ -63,8 +63,7 @@ def approximate_all_pairs_node_connectivity(
     pairs = list(iter_func(nbunch, 2))
     n_jobs = nxp.get_n_jobs()
     if get_chunks == "chunks":
-        num_in_chunk = max(min(len(pairs) // n_jobs, 10), 1)
-        pairs_chunks = nxp.chunks(pairs, num_in_chunk)
+        pairs_chunks = nxp.chunks(pairs, n_jobs)
     else:
         pairs_chunks = get_chunks(pairs)
 
--- a/nx_parallel/algorithms/connectivity/connectivity.py
+++ b/nx_parallel/algorithms/connectivity/connectivity.py
@@ -64,8 +64,7 @@ def all_pairs_node_connectivity(G, nbunch=None, flow_func=None, get_chunks="chun
     pairs = list(iter_func(nbunch, 2))
     n_jobs = nxp.get_n_jobs()
     if get_chunks == "chunks":
-        num_in_chunk = max(min(len(pairs) // n_jobs, 10), 1)
-        pairs_chunks = nxp.chunks(pairs, num_in_chunk)
+        pairs_chunks = nxp.chunks(pairs, n_jobs)
     else:
         pairs_chunks = get_chunks(pairs)
 
--- a/_nx_parallel/__init__.py
+++ b/_nx_parallel/__init__.py
@@ -21,14 +21,14 @@ def get_info():
                 },
             },
             "all_pairs_bellman_ford_path": {
-                "url": "https://github.com/networkx/nx-parallel/blob/main/nx_parallel/algorithms/shortest_paths/weighted.py#L212",
+                "url": "https://github.com/networkx/nx-parallel/blob/main/nx_parallel/algorithms/shortest_paths/weighted.py#L208",
                 "additional_docs": "The parallel implementation first divides the nodes into chunks and then creates a generator to lazily compute shortest paths for each node_chunk, and then employs joblib's `Parallel` function to execute these computations in parallel across `n_jobs` number of CPU cores.",
                 "additional_parameters": {
                     'get_chunks : str, function (default = "chunks")': "A function that takes in an iterable of all the nodes as input and returns an iterable `node_chunks`. The default chunking is done by slicing the `G.nodes` into `n_jobs` number of chunks."
                 },
             },
             "all_pairs_bellman_ford_path_length": {
-                "url": "https://github.com/networkx/nx-parallel/blob/main/nx_parallel/algorithms/shortest_paths/weighted.py#L168",
+                "url": "https://github.com/networkx/nx-parallel/blob/main/nx_parallel/algorithms/shortest_paths/weighted.py#L165",
                 "additional_docs": "The parallel implementation first divides the nodes into chunks and then creates a generator to lazily compute shortest paths lengths for each node in `node_chunk`, and then employs joblib's `Parallel` function to execute these computations in parallel across `n_jobs` number of CPU cores.",
                 "additional_parameters": {
                     'get_chunks : str, function (default = "chunks")': "A function that takes in an iterable of all the nodes as input and returns an iterable `node_chunks`. The default chunking is done by slicing the `G.nodes` into `n_jobs` number of chunks."
@@ -42,14 +42,14 @@ def get_info():
                 },
             },
             "all_pairs_dijkstra_path": {
-                "url": "https://github.com/networkx/nx-parallel/blob/main/nx_parallel/algorithms/shortest_paths/weighted.py#L124",
+                "url": "https://github.com/networkx/nx-parallel/blob/main/nx_parallel/algorithms/shortest_paths/weighted.py#L122",
                 "additional_docs": "The parallel implementation first divides the nodes into chunks and then creates a generator to lazily compute shortest paths for each `node_chunk`, and then employs joblib's `Parallel` function to execute these computations in parallel across `n_jobs` number of CPU cores.",
                 "additional_parameters": {
                     'get_chunks : str, function (default = "chunks")': "A function that takes in an iterable of all the nodes as input and returns an iterable `node_chunks`. The default chunking is done by slicing the `G.nodes` into `n_jobs` number of chunks."
                 },
             },
             "all_pairs_dijkstra_path_length": {
-                "url": "https://github.com/networkx/nx-parallel/blob/main/nx_parallel/algorithms/shortest_paths/weighted.py#L73",
+                "url": "https://github.com/networkx/nx-parallel/blob/main/nx_parallel/algorithms/shortest_paths/weighted.py#L72",
                 "additional_docs": "The parallel implementation first divides the nodes into chunks and then creates a generator to lazily compute shortest paths lengths for each node in `node_chunk`, and then employs joblib's `Parallel` function to execute these computations in parallel across `n_jobs` number of CPU cores.",
                 "additional_parameters": {
                     'get_chunks : str, function (default = "chunks")': "A function that takes in an iterable of all the nodes as input and returns an iterable `node_chunks`. The default chunking is done by slicing the `G.nodes` into `n_jobs` number of chunks."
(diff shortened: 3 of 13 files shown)


nx_parallel/algorithms/approximation/connectivity.py
```diff
@@ -63,8 +63,7 @@ def _process_pair_chunk(pairs_chunk):
     pairs = list(iter_func(nbunch, 2))
     n_jobs = nxp.get_n_jobs()
     if get_chunks == "chunks":
-        num_in_chunk = max(min(len(pairs) // n_jobs, 10), 1)
-        pairs_chunks = nxp.chunks(pairs, num_in_chunk)
```
Schefflera-Arboricola, line 67: This case is a bit different -- because here we don't want to have chunks with size greater than 10. This upper limit of 10 was set because with really large sized chunks this algorithm was not performing that well but with this upper limit it was showing good speed-ups. 

Maybe we can add a new `max_chunk_size` parameter in the `chunks` function and modify the logic accordingly. LMKWYT.

nx_parallel/algorithms/connectivity/connectivity.py
```diff
@@ -64,8 +64,7 @@ def _process_pair_chunk(pairs_chunk):
     pairs = list(iter_func(nbunch, 2))
     n_jobs = nxp.get_n_jobs()
     if get_chunks == "chunks":
-        num_in_chunk = max(min(len(pairs) // n_jobs, 10), 1)
-        pairs_chunks = nxp.chunks(pairs, num_in_chunk)
```
Schefflera-Arboricola, line 68: same here

Schefflera-Arboricola (commented):
Thanks @akshitasure12 ! Nice work :)

I just have one comment below. LMKWYT.

nx_parallel/algorithms/approximation/connectivity.py
akshitasure12: Sure, I've updated it! 
:)

nx_parallel/algorithms/connectivity/connectivity.py
```diff
@@ -65,7 +65,8 @@ def _process_pair_chunk(pairs_chunk):
     n_jobs = nxp.get_n_jobs()
     if get_chunks == "chunks":
         num_in_chunk = max(min(len(pairs) // n_jobs, 10), 1)
-        pairs_chunks = nxp.chunks(pairs, num_in_chunk)
+        max_chunk_size = 10 if num_in_chunk == 10 else None
```
Schefflera-Arboricola, line 68: I think it would better to have this logic/check(i.e. if balanced chunking gives chunks of size greater than max_chunk_size then run the new logic-- that returns chunks of size at most max_chunk_size) in the `chunks()` function itself.

Also please consider documenting the behaviour in `chunks`'s docstring as well.

nx_parallel/utils/chunk.py
```diff
@@ -6,15 +6,23 @@
...
+    if max_chunk_size is not None:
+        for chunk in itertools.batched(iterable, max_chunk_size):
+            yield chunk
+    else:
+        k, m = divmod(len(iterable), n_chunks)
+        it = iter(iterable)
+        for _ in range(n_chunks):
+            chunk_size = k + (1 if m > 0 else 0)
+            m -= 1
+            yield tuple(itertools.islice(it, chunk_size))
```
Schefflera-Arboricola, line 25: no need for an explicit `else`
```suggestion
    k, m = divmod(len(iterable), n_chunks)
    it = iter(iterable)
    for _ in range(n_chunks):
        chunk_size = k + (1 if m > 0 else 0)
        m -= 1
        yield tuple(itertools.islice(it, chunk_size))
```

Also could you please add a test for the new `max_chunk_size` [here](https://github.com/networkx/nx-parallel/blob/22e4c3cd6541528dff28c991821d2909fc105b1d/nx_parallel/utils/tests/test_chunk.py#L46)?

nx_parallel/algorithms/approximation/connectivity.py
```diff
@@ -64,7 +64,8 @@ def _process_pair_chunk(pairs_chunk):
     n_jobs = nxp.get_n_jobs()
     if get_chunks == "chunks":
         num_in_chunk = max(min(len(pairs) // n_jobs, 10), 1)
```
Schefflera-Arboricola, line 66: same here-- num_in_chunks should be decided inside `chunks()`

nx_parallel/algorithms/connectivity/connectivity.py
akshitasure12: Building on that, if the num_in_chunks is determined internally by chunks(), do we still need to pass max_chunk_size as a parameter?

Instead, we could maintain a list of functions or algorithms where limiting the chunk size provides better performance. The chunks() function could then check this internally and enforce max_chunk_size only when necessary. Let me know what you think!

nx_parallel/utils/chunk.py
akshitasure12: > no need for an explicit `else`

Removing the else would result in double yields, am I missing something here?

nx_parallel/algorithms/connectivity/connectivity.py
Schefflera-Arboricola: > The chunks() function could then check this internally and enforce max_chunk_size only when necessary

how will `chunks()` know in which xyz function it's being called from?

nx_parallel/utils/chunk.py
```diff
@@ -6,9 +6,19 @@
...
+    Yield chunks by splitting the iterable either into chunks of size at
+    most `max_chunk_size`, or into exactly `n_chunks` balanced chunks if
+    `max_chunk_size` is None.
+    """
     iterable = list(iterable)
+    if max_chunk_size:
+        num_in_chunk = max(min(len(iterable) // n_chunks, max_chunk_size), 1)
+        if num_in_chunk == max_chunk_size:
+            yield from itertools.batched(iterable, max_chunk_size)
+            return
```
akshitasure12, line 20: I added a return here because if `num_in_chunk` is equal to `max_chunk_size`, without the return, it would execute both the yield statements. Let me know if any other changes need to be done!

nx_parallel/utils/chunk.py
```diff
@@ -6,9 +6,19 @@
...
 
 
-def chunks(iterable, n_chunks):
-    """Yield exactly `n_chunks` chunks from `iterable`, balancing the chunk sizes."""
+def chunks(iterable, n_chunks, max_chunk_size=None):
+    """
+    Yield chunks by splitting the iterable either into chunks of size at
+    most `max_chunk_size`, or into exactly `n_chunks` balanced chunks if
+    `max_chunk_size` is None.
+    """
```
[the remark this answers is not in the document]
akshitasure12: Sure, I'll do that!

nx_parallel/utils/chunk.py
```diff
@@ -6,9 +6,42 @@
...
-def chunks(iterable, n_chunks):
-    """Yield exactly `n_chunks` chunks from `iterable`, balancing the chunk sizes."""
+def chunks(iterable, n_chunks, *, max_chunk_size=None):
+    """Yield chunks from input iterable.
+
+    - If `max_chunk_size` is None (default), the iterable
+    is split into exactly `n_chunks` balanced chunks.
+    - If `max_chunk_size` is specified, it overrides `n_chunks`
+    and the iterable is split into chunks of size at most
+    `max_chunk_size`.
```
Schefflera-Arboricola, line 16: it only overrides if chunk sizes in the balanced chunking exceeds the max_chunk_size value. 

also "balanced chunks" is not a well-defined term-- something like "equally sized chunks" might be more understandable.

nx_parallel/utils/chunk.py
```diff
@@ -6,9 +6,42 @@
...
+
+    Parameters
+    ----------
+    iterable : Iterable
+        An iterable of inputs to be divided.
+    n_chunks : int
+        The number of chunks the iterable is divided into. Ignored
+        when `max_chunk_size` is specified.
```
Schefflera-Arboricola, line 24: not always ignored-- see above comment

nx_parallel/utils/chunk.py
```diff
@@ -6,9 +6,42 @@
...
+    >>> import nx_parallel as nxp
+    >>> data = list(range(10))
+    >>> list(nxp.chunks(data, 3))
+    [(0, 1, 2, 3), (4, 5, 6), (7, 8, 9)]
+    >>> list(nxp.chunks(data, 3, max_chunk_size=2))
+    [(0, 1), (2, 3), (4, 5), (6, 7), (8, 9)]
+    """
     iterable = list(iterable)
+    if max_chunk_size:
+        num_in_chunk = max(min(len(iterable) // n_chunks, max_chunk_size), 1)
```
Schefflera-Arboricola, line 40: we can simplify and move this computation outside the `if max_chunk_size` condition so that we don't do it again in line 45. Also it would be better to have more meaningful variable names than `k` and `m`.

nx_parallel/utils/chunk.py
```diff
@@ -6,14 +6,46 @@
...
 
 
-def chunks(iterable, n_chunks):
-    """Yield exactly `n_chunks` chunks from `iterable`, balancing the chunk sizes."""
+def chunks(iterable, n_chunks, *, max_chunk_size=None):
+    """Yield chunks from input iterable.
+
+    - If `max_chunk_size` is None (default), the iterable is split into
+    exactly `n_chunks` equally sized chunks.
+    - If `max_chunk_size` is specified and the calculated `base_chunk_size`
```
Schefflera-Arboricola, line 14: please define what's `base_chunk_size` before using it-- or not mention it here at all since it is not an exposed variable.

nx_parallel/utils/tests/test_chunk.py
```diff
@@ -44,6 +44,10 @@ def test_chunks():
     chunks_list = list(nxp.chunks(data, 5))
     assert chunks_list == [(0, 1), (2, 3), (4, 5), (6, 7), (8, 9)]
 
+    # Test chunking with a maximum chunk size
+    chunks_list = list(nxp.chunks(data, 2, max_chunk_size=3))
+    assert chunks_list == [(0, 1, 2), (3, 4, 5), (6, 7, 8), (9,)]
+
```
Schefflera-Arboricola, line 50: consider adding a test for when `max_chunk_size>base_chunk_size`

Schefflera-Arboricola (commented):
this looks almost done now!

nx_parallel/utils/chunk.py
```diff
@@ -6,14 +6,48 @@
...
-    """Yield exactly `n_chunks` chunks from `iterable`, balancing the chunk sizes."""
+def chunks(iterable, n_chunks, *, max_chunk_size=None):
+    """Yield chunks from input iterable.
+
+    - If `max_chunk_size` is None (default), the iterable is split into
+    exactly `n_chunks` equally sized chunks.
+    - If `max_chunk_size` is specified and the default split would create
+    chunks larger than this size, the iterable is instead divided into
+    smaller chunks, each containing at most `max_chunk_size` items.
+    - In any other case, it falls back to the default behavior.
```
Schefflera-Arboricola, line 17:
```suggestion
```

Schefflera-Arboricola (changes requested):

dschult:
I can't recreate the test failures locally... even with trying to use the CI workflow commands.  
I'm tempted to just merge this because the errors are only square clustering.

Am I missing anything?

PR #116, opened 2025-06-05
Drop support for Python 3.11

dschult:
We'd like to use `itertools.batched` in our chunking tools and that is only added for Python 3.12+.
So we could put together our own version of batched and change it next Fall when we drop Python 3.11 support. Or we can drop the support for Python 3.11 now and use the official `itertools.batched` tool. As discussed in the meeting this PR drops the support for Python 3.11.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,7 +89,7 @@ attr = 'nx_parallel.__version__'
 
 [tool.ruff]
 line-length = 88
-target-version = 'py311'
+target-version = 'py312'
 
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ['I', 'F403', 'F401']


Schefflera-Arboricola (approved):

pyproject.toml
```diff
@@ -90,7 +89,7 @@ attr = 'nx_parallel.__version__'
 
 [tool.ruff]
 line-length = 88
-target-version = 'py311'
+target-version = 'py312'
```
Schefflera-Arboricola, line 92: Not sure if this gets used?

Schefflera-Arboricola: https://github.com/networkx/networkx/pull/8102 for reference

Issue #79, opened 2024-08-24
Addressing Slower Parallel Implementations in nx-parallel Compared to NetworkX


Schefflera-Arboricola:
## Description:

Need to address the performance of certain algorithms in the `nx-parallel` whose parallel implementations are currently slower than their corresponding implementations in NetworkX. Some of these algorithms include:

- `v_structures` in https://github.com/networkx/nx-parallel/pull/74
- `colliders` in https://github.com/networkx/nx-parallel/pull/74
- `global_reaching_centrality` in https://github.com/networkx/nx-parallel/pull/44
- `local_reaching_centrality` in https://github.com/networkx/nx-parallel/pull/44
- `average_clustering`
- `approximate_average_clustering`
- [`number_of_isolates`](https://github.com/networkx/nx-parallel/blob/main/nx_parallel/algorithms/isolate.py#L8)

## Issues to Discuss:

1. **Performance Comparison**: What should we do with these functions that show slower performance in their parallel implementations? Should we consider reverting to or recommending the use of NetworkX's implementation in these cases?
   
2. **Optimization**: Is there a more optimized way that would improve the performance of these parallel implementations? And will that be way better the performance for all the algorithms? And how can we give the control of changing these "ways" to the end user?

3. **Functionality**: There are also some graph algorithms that cannot be parallelized due to their fundamentally non-parallelizable nature. How should we handle such algorithms in nx-parallel? Should we provide a clear indication in the documentation or in the code itself? Or not add them to nx-parallel at all?

## Next Steps:

- Gather insights and suggestions from the community.
- Determine whether to optimize, remove, or replace the slower parallel implementations.
- Identify any non-parallelizable functions and decide on how to best handle them.

Thank you :)

PR #117, opened 2025-06-07
added parallel implementations for `number_` algorithms

akshitasure12:
added parallel implementations for 
- [number_attracting_components](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.number_attracting_components.html)
- [number_connected_components](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.number_connected_components.html)
- [number_strongly_connected_components](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.number_strongly_connected_components.html)
- [number_weakly_connected_components](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.number_weakly_connected_components.html)

These algorithms involve lightweight computations (e.g., simple summations), limiting their performance. Their behaviour should be analysed in context of PR #79.


akshitasure12:
in the recent commit, I attempted to add support for the `should_run` check .

as per the comment mentioned [here](https://github.com/networkx/networkx/pull/7257#issue-2106598468), `should_run` is only triggered when a NetworkX graph would be converted to a backend graph. However, I tried the following:
```
>>> import networkx as nx
>>> G = nx.empty_graph(5)
>>> nx.config.backends.parallel.active = True
>>> nx.number_of_isolates(G)
5
```
can you help me understand why `should_run` is not getting triggered in the following scenario?

also, are we planning to support `can_run`?

nx_parallel/__init__.py
```diff
@@ -1,6 +1,7 @@
 from .utils.decorators import _configure_if_nx_active
 from .utils import *
 from .algorithms import *
+from .algorithms import components
```
dschult, line 4: This shouldn't be needed if we import inside `algorithms/__init__.py`, right?  I think this will provide `nx_parallel.components` instead of `nx_parallel.algorithms.components`.

nx_parallel/algorithms/components/attracting.py
```diff
@@ -0,0 +1,35 @@
+from networkx.algorithms.components.attracting import attracting_components
```
dschult, line 1: This can be `from networkx import attracting_components`. 
And that is the standard way to import within NetworkX. (It is a "flat" namespace to a large extent). There are a few places where it isn't flat: Many of the `bipartite` algorithms have the same names as the regular functions, so we put them into a sub-namespace: `nx.bipartite.closeness_centrality()`.  Of course people **can** call them with the full pathname. But we encourage people to use the shorter syntax.

akshitasure12:
> I guess to make any speedup we will need to parallelize `attracting_components`. Right?

Yes-- these functions are similar to that of `number_of_isolates`. To observe any speedups, we would need to parallelize them; however, functions like `attracting_components`, `connected_components`, etc., are inherently sequential.

I plan to add the `should_run` flag-- `should_skip_parallel` (ref. PR #123)-- so that the parallel backend is bypassed for these functions.

akshitasure12:
@dschult, this PR is ready!

dschult (approved):

PR #119, opened 2025-06-09
parallel implementation of `is_reachable` using mem-mapping

akshitasure12:
- adding a numpy adjacency matrix based representation of `is_reachable()` yielding better speedups (ref [comment](https://github.com/networkx/nx-parallel/pull/114#issuecomment-2953902805)).
- utilising shared memory to avoid passing `G` across multiple cores via joblib's [memmapping](https://joblib.readthedocs.io/en/latest/parallel.html#working-with-numerical-data-in-shared-memory-memmapping).

### DIFF
--- a/nx_parallel/algorithms/tournament.py
+++ b/nx_parallel/algorithms/tournament.py
@@ -11,7 +11,7 @@ __all__ = [
 ]
 
 
-@nxp._configure_if_nx_active()
+@nxp._configure_if_nx_active(should_run=nxp.should_skip_parallel)
 def is_reachable(G, s, t, get_chunks="chunks"):
     """The function parallelizes the calculation of two
     neighborhoods of vertices in `G` and checks closure conditions for each


Schefflera-Arboricola:
something on the similar lines(in terms of enhancing tournament algorithms)-- we can also consider taking this [reusing the pool of workers](https://joblib.readthedocs.io/en/latest/parallel.html#reusing-a-pool-of-workers) approach maybe for tournament_is_strongly_connected (in line 80) -- and that might show some more improvements in terms of speedups-- but all our functions are currently wrapped in a `_configure_if_nx_active`-- which is essentially a  `joblib.parallel_config` context-- so this would require us to change how `_configure_if_nx_active` works currently.

PS: this doesn't have to be handled in this PR-- just something we can think about in the future.

nx_parallel/algorithms/tournament.py
```diff
@@ -25,33 +27,50 @@ def is_reachable(G, s, t, get_chunks="chunks"):
...
+                or any(adjM[v, z] and adjM[z, x] for z in node_indices)
             }
-            tnc.append(not (is_closed(G, S) and s in S and t not in S))
+            tnc.append(not (is_closed(adjM, S) and s_ind in S and t_ind not in S))
         return all(tnc)
 
-    def is_closed(G, nodes):
-        return all(v in G[u] for u in set(G) - nodes for v in nodes)
+    def is_closed(adjM, nodes):
+        other_nodes = set(range(adjM.shape[0])) - nodes
```
dschult, line 46: maybe pass in `node_indices` so we can reuse that computation inside the loop.

nx_parallel/algorithms/tournament.py
```diff
@@ -25,33 +27,50 @@ def is_reachable(G, s, t, get_chunks="chunks"):
...
         tnc = []
         for v in chunk:
             S = {
                 x
-                for x in G
-                if x == v or x in G[v] or any(is_path(G, [v, z, x]) for z in G)
+                for x in node_indices
+                if x == v
+                or adjM[v, x]
+                or any(adjM[v, z] and adjM[z, x] for z in node_indices)
```
dschult, line 40: Can this be sped up with matrix multiplication? It looks like it is `adjM[v,:] @ adjM[:, x]` but only computing with indexing that are in `node_indices`.  We can limit it to only those indices using array indexing. Something like:
```python
adjM[v,node_indices] @ adjM[node_indices, x]
```

nx_parallel/algorithms/tournament.py
```diff
@@ -25,33 +27,50 @@ def is_reachable(G, s, t, get_chunks="chunks"):
...
             }
-            tnc.append(not (is_closed(G, S) and s in S and t not in S))
+            tnc.append(not (is_closed(adjM, S) and s_ind in S and t_ind not in S))
         return all(tnc)
 
-    def is_closed(G, nodes):
-        return all(v in G[u] for u in set(G) - nodes for v in nodes)
+    def is_closed(adjM, nodes):
```
dschult, line 45: Try checking the speed with and without this function. That is, inline the contents of the function into line 42 so Python can avoid a function call.  

Another possible speed-up could be to put the simple checks first in the compound logical statement in line 42. That should avoid the `is_closed` code being run at all unless s in S and t not in S.

Do we know where this is taking most of its time?  I was just guessing for these suggestions based on the for loop structure.  Using the ipython tool `%prun` you might be able to tell where the code spends the most time -- but I don't know how that works with parallel joblib. Maybe try it with 1 cpu or with the parallel code part removed.

akshitasure12: > Try checking the speed with and without this function. That is, inline the contents of the function into line 42 so Python can avoid a function call.

I tried adding this logic to line 42, but it didn’t result in any noticeable speedups. So I’ve decided to keep it as a separate function for better readability. Pls let me know if you think otherwise.

> Do we know where this is taking most of its time? I was just guessing for these suggestions based on the for loop structure. Using the ipython tool `%prun` you might be able to tell where the code spends the most time -- but I don't know how that works with parallel joblib. Maybe try it with 1 cpu or with the parallel code part removed.

 I removed the parallel execution and ran `%prun` using the following input:
```python
G = nx.algorithms.tournament.random_tournament(1600)
s, t = 0, 1599
```
A summary of the output:
```bash
ncalls      tottime  percall  cumtime  percall filename:lineno(function)
     3/1     0.000    0.000    5.383    5.383   {built-in method builtins.exec}
       1     0.000    0.000    5.383    5.383   <string>:1(<module>)
       1     0.000    0.000    5.383    5.383   decorators.py:17(wrapper)
       1     0.000    0.000    5.383    5.383   tournament.py:14(is_reachable)
       1     0.728    0.728    4.324    4.324   tournament.py:30(two_neighborhood_close)
 1279201     0.296    0.000    2.851    0.000   {built-in method builtins.any}
 6404724     0.878    0.000    2.674    0.000   tournament.py:40(<genexpr>)
 10230374    2.418    0.000    2.418    0.000   memmap.py:357(__getitem__)
       1     0.000    0.000    1.056    1.056   decorators.py:783(func)
       1     0.000    0.000    1.056    1.056   <class 'networkx.utils.decorators.argmap'> compilation 21:1(argmap_to_numpy_array_18)
       1     0.000    0.000    1.042    1.042   backends.py:538(_call_if_any_backends_installed)
       1     0.006    0.006    1.042    1.042   backends.py:1460(_call_with_backend)
       1     0.484    0.484    1.035    1.035   convert_matrix.py:881(to_numpy_array)
```
It is evident from this that the bulk of time is spent computing two-hop neighbors and the memory access via `memmap`. I've tried other alternatives you mentioned, but this seems to perform the best.

akshitasure12:
Apologies for the foresight! I was able to figure it out from the documentation. Thank you @dschult!

dschult:
> Do these benchmarks imply that the process of converting a graph into a numpy matrix is more expensive than that of passing the huge graph to each core? Because n_jobs is None by default, for these benchmarks, I'm unsure how we are seeing the above speedups.

Which version of config is used when `n_jobs is None`? That is, how many jobs are you actually running?  The last benchmark with parallel w/ python vs sequential python looks like parallel is about 16 times faster.  So that is either 16 cores or 4 cores with quadratic speed-up.  I don't see any quadratic speedup looking at the code. But it is possible I am missing something.

Can you point me to your benchmarking code? Is it already merged in nx-parallel, or is it only on your local machine?

akshitasure12:
> That is, how many jobs are you actually running?

I didn’t modify the number of jobs anywhere in the codebase, so by default, `n_jobs=None` is being used by joblib which would mean 1 core. I also verified this by checking the activity monitor while running the benchmark — the CPU usage confirmed that only a single core was active during the run.

> Can you point me to your benchmarking code? Is it already merged in nx-parallel, or is it only on your local machine?

The local changes I made were:
- replacing
   ```
    _ = nx.tournament.is_reachable(G, 1, num_nodes, backend=backend)
   ```
   with:
   ```
   _ = nx.tournament.is_reachable(G, 0, num_nodes - 1, backend=backend)
   ```
   to prevent the target node from going out of bounds.
- adding "networkx" to the list of backends in common.py.

akshitasure12:
I ran the timing script on both numpy implementation and the pure python implementation.
Numpy:
```bash
200
0.16524791717529297
0.0043315887451171875
Finished <function is_reachable at 0x1037ba980>
400
0.04875922203063965
0.017219066619873047
Finished <function is_reachable at 0x1037ba980>
800
0.1309349536895752
0.06830096244812012
Finished <function is_reachable at 0x1037ba980>
1600
0.4890291690826416
0.31351304054260254
Finished <function is_reachable at 0x1037ba980>
is_reachable
```
Python:
```bash
200
0.2303180694580078
0.004209995269775391
Finished <function is_reachable at 0x105c5a980>
400
0.056372880935668945
0.016617774963378906
Finished <function is_reachable at 0x105c5a980>
800
0.30538177490234375
0.06642794609069824
Finished <function is_reachable at 0x105c5a980>
1600
1.9619712829589844
0.3054389953613281
Finished <function is_reachable at 0x105c5a980>
is_reachable
```
These timings clearly show that the NumPy version outperforms the pure Python parallel version as the graph size grows. This is mainly because converting the graph to an adjacency matrix is a one-time cost, while passing the entire Graph object across multiple cores in the pure Python version repeatedly triggers expensive pickling and unpickling. The current implementation avoids this by storing the matrix on disk and allowing each worker to read it directly (using mem-mapping). I’m happy with this approach instead of converting it to [pure Python parallel code](https://github.com/networkx/nx-parallel/pull/119#issuecomment-3091992580), but please let me know if you have any other suggestions.

Issue #115, opened 2025-06-05
Use context manager in tests instead of modifying the global configurations


Schefflera-Arboricola:
> How does one use the pytest.mark.order? Set it to 2 for tests that set config settings? And leave it off otherwise? Why use mark.order rather than resetting the values after the test finishes? Is it that a failed test will not reset the config? Can we use a context manager to reset the config settings?

I agree that using a context manager would make more sense for running tests because then the tests would be more independent of each other-- like you mentioned the case when the first test fails then the configs may or may not be what the second test would expect.

_Originally posted by @Schefflera-Arboricola in https://github.com/networkx/nx-parallel/pull/113#discussion_r2123174554_

PR #120, opened 2025-06-23
Remove test order dependency using context managers

akshitasure12:
Handles #115 

- Replaced manual config resets with context managers to isolate config changes.
- Removed `pytest.mark.order` decorator from `test_get_n_jobs()` and `test_config_init()` since test order is no longer needed.

### DIFF
--- a/nx_parallel/utils/tests/test_chunk.py
+++ b/nx_parallel/utils/tests/test_chunk.py
@@ -4,7 +4,6 @@ import networkx as nx
 import nx_parallel as nxp
 
 
-@pytest.mark.order(2)
 def test_get_n_jobs():
     """Test for various scenarios in `get_n_jobs`."""
     # Test with no n_jobs (default)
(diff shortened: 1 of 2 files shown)


dschult (approved):
This looks like a straightforward simplification to tests.
It takes advantage of context managers for setting of config variables.

Tests are passing. Looks right.

nx_parallel/utils/tests/test_chunk.py
```diff
@@ -17,17 +16,19 @@ def test_get_n_jobs():
...
-        # restore nx.config
-        nx.config.backends.parallel.active = False
-        nx.config.backends.parallel.n_jobs = None
+        with parallel_config(n_jobs=3):
+            assert nxp.get_n_jobs() == 3
+
+        # Test with nx-parallel's context
+        nxp_config = nx.config.backends.parallel
+
+        with nxp_config(active=True, n_jobs=5):
```
Schefflera-Arboricola, line 29: no need to create a variable

```suggestion
        with nx.config.backends.parallel(active=True, n_jobs=5):
```

PR #124, opened 2025-07-18
[WIP] : parallel implementation of harmonic centrality

akshitasure12:
- parallel version is extremely slow compared to the serial.
- max performance observed for:
   n_jobs = 2
   image
   n_jobs = 3
   image
- Should consider adding a `should_run`?


akshitasure12:
Notable Optimisation:
In the original NetworkX implementation of harmonic centrality, the code computes `nbunch.intersection(dist)` at every iteration. This introduces unnecessary memory overhead, and the computation is redundant. I replaced 
```py
for v in sources:
    dist = spl(v)
    for u in nbunch.intersection(dist):
         d = dist[u]
         if d == 0:  # handle u == v and edges with 0 weight
             continue
         centrality[v if transposed else u] += 1 / d
```
with:
```py
for src in chunk:
    dist = spl(src)
    for tgt, dst in dist.items():
         if tgt in nbunch and dst != 0:
             node = src if transposed else tgt
             hc[node] = hc.get(node, 0) + 1/dst
```
I avoided the explicit set creation by directly iterating over `dist.items()`. Let me know your thoughts on this!

akshitasure12:
Yes -- I agree that the naming could be improved. I had initially used `src` and `tgt` for a sense of directionality maybe, but I can switch to `u` and `v` to be more consistent with the convention. Also, I see how dist vs dst can be confusing -- I’ll rename them to something clearer, suggested above.
Thank you :)

dschult:
I found a section in the joblib docs that talks about how joblib transfers large objects to the processes.
https://joblib.readthedocs.io/en/latest/parallel.html#serialization-processes
There are some StackOverflow questions and answers about why parallelization can be slow.
https://stackoverflow.com/questions/65026499/why-does-joblib-parallel-execution-make-runtime-much-slower

akshitasure12:
> I found a section in the joblib docs that talks about how joblib transfers large objects to the processes.
> https://joblib.readthedocs.io/en/latest/parallel.html#serialization-processes

The challenge with using `pickle` is that it cannot serialize local objects, which means we’d need to make `_process_chunk` a global function (ref. [StackOverflow thread](https://stackoverflow.com/questions/72766345/attributeerror-cant-pickle-local-object-in-multiprocessing)). 

Also, in our previous meeting-- we questioned why `betweenness_centrality` benefits from parallelism while `harmonic_centrality` does not. The reason is that after computing single-source shortest paths, `betweenness_centrality` re-traverses the shortest path to back-propagate scores. This means each source requires two passes-- one to find paths and another for score propagation-- making it computationally expensive. In contrast, `harmonic_centrality` involves only lightweight binary operations after computing shortest paths, so parallelism offers little gain.

Given this, I’m considering adding the `should_run` parameter here after the merge. Let me know your thoughts!

dschult:
This example shows speedups for sparse networks. (p=0.2)

This makes me start to rethink our benchmark parameters.
We are timing for p: 0.2, 0.4, 0.6, 0.8, 1.0 which spreads out equally among the p values 0 and 1. But networks are almost always sparse (otherwise we wouldn't call them networks -- we would just track everyone to everyone contact). We should probably be using much larger numbers of nodes and much smaller density values.  But that is a different discussion. I will open an issue for that.

I guess here we need to decide what `should_run` should return. We should run when the sparsity is high, but not when it is low. We could do a more detailed set of timings to find the right criteria, right?  What do you suggest for should_run?

nx_parallel/algorithms/centrality/harmonic.py
```diff
@@ -0,0 +1,64 @@
...
+                if d_uv != 0 and u in nbunch:
+                    node = v if transposed else u
+                    hc[node] = (hc[node] + 1 / d_uv) if node in hc else (1 / d_uv)
+        return hc
+
+    if hasattr(G, "graph_object"):
+        G = G.graph_object
+
+    nbunch = set(G.nbunch_iter(nbunch) if nbunch is not None else G.nodes)
+    sources = set(G.nbunch_iter(sources) if sources is not None else G.nodes)
```
dschult, line 40: The logic for handling `nbunch is None` is already included in `G.nbunch_iter`. And we are creating a set out of the iterator anyway, so I think it's just as performant to use: `nbunch = set(G.nbunch_iter(nbunch))` and similarly for `sources`.

nx_parallel/algorithms/centrality/harmonic.py
```diff
@@ -0,0 +1,64 @@
...
+
+    transposed = False
```
dschult, line 42: Do you use transposed somewhere? Or can that be removed now?

dschult (approved):
Yes -- this one looks ready.

nx_parallel/algorithms/centrality/harmonic.py
```diff
@@ -0,0 +1,64 @@
+from joblib import Parallel, delayed
+import networkx as nx
+import nx_parallel as nxp
+from functools import partial
+
+__all__ = ["harmonic_centrality"]
+
+
+@nxp._configure_if_nx_active(should_run=nxp.should_run_if_sparse)
```
dschult, line 9: I've got a question on the density-based-threshold version of `should_run`.
How do we specify that the threshold for this function should be 0.3? What is we want to change the threshold to 0.1? I see that the function itself has a default threshold of 0.3. But how do we decorate with a threshold that is specific for this function?

Issue #125, opened 2025-07-19
using a setup function in benchmarks


akshitasure12:
I expect that we will need many different types of graphs for the algorithms. So the `get_graph` May not be able to be used very often. At least that is what happens with tests and pytest.  We create a test class and put some graphs that could be used by lots of tests, but then they end up only being used by a couple of the tests and the rest construct their own graph. So we may want to create a different setup function for many of the timing functions. I think we’ll have to wait to see what is most useful.  Keeping it as a common function is fine with me. Could we name it something like: `cached_gnp_random_graph`?  

Does caching  work well with floating point inputs like `p`? Nice binary values like 0.25 should work, but round-off can occur with 0.3, e.g.  I guess if the roundoff is the same for each call of the function we’re ok.

In general, caching is nice for time consuming parts that either take a lot of time, or get repeated many times. We’ll see as we get more examples whether the timing is taking a long time and how much of that time is constructing the graphs. Let’s leave it in for now.

_Originally posted by @dschult in https://github.com/networkx/nx-parallel/issues/24#issuecomment-1828078576_

dschult:
Yes -- I am used to a setup function being used with `asv` for each timing script. Sometimes you can use the same setup for multiple functions, but I've often seen a different setup for each timing. 

In this case the tournament graphs are special so should have a setup to avoid building the graph inside the function. And we do not want to use any kind of centralized setup function. At least as far as I can imagine.

PR #126, opened 2025-07-20
Refactor ASV benchmarks to use setup functions

akshitasure12:
Handles issue #125 

### Summary of changes introduced:
- Introduced a `setup()` function in each benchmark class to handle graph creation and, where necessary, convert the graph to a backend-specific format (e.g., ParallelGraph for `all_pairs_node_connectivity`).
- Used separate attributes (`self.G` for unweighted and `self.G_weighted` for weighted graphs) when a benchmark class required both types.
- Centralized the random seed (`seed = 42`) in common.py for consistent graph generation across all benchmarks.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -20,9 +20,12 @@ These asv benchmarks are not just good to see how parallel implementations are i
 
 ## Structure of benchmarks
 
-- Each `bench_` file corresponds to a folder/file in the [networkx/algorithms](https://github.com/networkx/networkx/tree/main/networkx/algorithms) directory in NetworkX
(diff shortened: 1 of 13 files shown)


dschult:
It seems like we shouldn't be including the graph creation time when timing an algorithm. Otherwise our ratios of times don't show how much speed-up the algorithm has. For example, ratio = (0.5 + 0.1)/(0.5+0.3) = 0.75 but 0.1/0.3 = 0.33.

When graph creation times are small it doesn't matter, but we probably want to be able to change the size of the graphs without changing the benchmark code. Maybe I am missing something???

PR #127, opened 2025-07-21
Parallelise link prediction algorithms

akshitasure12:
This PR introduces a parallel implementation of the `_apply_prediction` helper, which enables parallel execution of multiple link prediction algorithms that rely on this internal function.

The following link prediction algorithms are parallelized:
- [resource_allocation_index](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.link_prediction.resource_allocation_index.html)
- [jaccard_coefficient](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.link_prediction.jaccard_coefficient.html)
- [adamic_adar_index](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.link_prediction.adamic_adar_index.html)
- [preferential_attachment](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.link_prediction.preferential_attachment.html)

Results from the timing script show that `resource_allocation_index`, `jaccard_coefficient`, and `adamic_adar_index` achieve speedups of 2–5× compared to the original sequential implementations. Note that `preferential_attachment` does not show significant speedup, as its predict function is computationally trivial (a single multiplication). 

### Differences from NetworkX’s original `_apply_prediction`:
If `ebunch` is empty, the function immediately returns an empty iterator, skipping unnecessary parallel setup and overhead. When a graph is fully connected (e.g., edge_probability = 1), `ebunch` defaults to `None`; `nx.non_edges(G)` is used to generate pairs. Since `nx.non_edges(G)` returns an empty generator, this check avoids extra work.
```py
if not ebunch:
        return iter([])
```


nx_parallel/algorithms/link_prediction.py
```diff
@@ -0,0 +1,296 @@
...
+    n_jobs = nxp.get_n_jobs()
+
+    if get_chunks == "chunks":
+        pairs_chunks = nxp.chunks(ebunch, n_jobs)
+    else:
+        pairs_chunks = get_chunks(ebunch)
+
+    results = Parallel()(delayed(_process_pair_chunk)(chunk) for chunk in pairs_chunks)
+
+    return sorted(list(itertools.chain.from_iterable(results)))
```
dschult, line 51: I think this `sorted` call will raise errors on nodes that can't be compared, e.g. when some nodes are integers and others are strings.  In general, nodes are not sortable.  Do we need to sort the results here? Can they be in any order?

dschult (commented):
This looks good on a first read-through. I have two minor questions below.

nx_parallel/algorithms/link_prediction.py
akshitasure12: Thank you for pointing that out! The intention behind using `sorted` here was to ensure that `test_get_chunks()` passes. Without sorting, the test would fail because it compares the output of the function run with random chunking against the default nxp chunking. While the results are identical in content, they differ in order. I'll look into other ways to make this work :)

benchmarks/benchmarks/bench_link_prediction.py
```diff
@@ -0,0 +1,43 @@
...
+    num_nodes,
+    edge_prob,
+    get_cached_gnp_random_graph,
+    Benchmark,
+)
+import networkx as nx
+
+
+class LinkPrediction(Benchmark):
+    params = [(backends), (num_nodes), (edge_prob)]
```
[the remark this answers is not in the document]
akshitasure12: I don’t think the `param_names` would make much sense without params, since the names just label the parameter values.

akshitasure12:
Pain point:
You cannot return a generator inside `_process_chunks` because it isn’t serializable, so we would have to return a list instead. When we return a list, every exception happens immediately during its creation. In the current test:
```py
pytest.raises(nx.NetworkXAlgorithmError, list, self.func(G, [(0, 1)]))
```
`self.func` is evaluated before `pytest.raises` can catch the exception, since it’s run outside the context. So I think it would make more sense to rewrite this test using the context manager form, which ensures any exception raised during evaluation is caught properly:
```py
with pytest.raises(nx.NetworkXAlgorithmError):
    list(self.func(G, [(0, 1)]))
```
This change would mean updating the NetworkX test to use the context manager form instead. Let me know what you think!

dschult (approved):
I think this is ready now!   Can we remove the WIP from the title yet?
Are there other things this is waiting on?

I approve this PR

akshitasure12:
> Are there other things this is waiting on? Like maybe the get_chunk PR #129

Yes, I'm just waiting for PR #129 to be merged, which should ensure these tests pass.

akshitasure12:
Thank you, @dschult! I’ve rebased, and everything looks good to me now :)

PR #128, opened 2025-07-24
refactor `test_get_functions_with_get_chunks`

akshitasure12:
This PR refactors the `test_get_functions_with_get_chunks` test to dynamically use the `ALGORITHMS` from interface.py instead of relying on a hardcoded expected set. It strips module prefixes for algorithms like `connectivity.all_pairs_node_connectivity`.

### DIFF
--- a/nx_parallel/tests/test_get_chunks.py
+++ b/nx_parallel/tests/test_get_chunks.py
@@ -31,7 +31,7 @@ def get_functions_with_get_chunks(ignore_funcs=[], package_name="nx_parallel"):
 
 
 def test_get_functions_with_get_chunks():
-    assert set(get_functions_with_get_chunks()) == set(ALGORITHMS)
+    assert get_functions_with_get_chunks() == ALGORITHMS
 
 
 ignore_funcs = [
--- a/nx_parallel/tests/test_get_chunks.py
+++ b/nx_parallel/tests/test_get_chunks.py
@@ -31,7 +31,7 @@ def get_functions_with_get_chunks(ignore_funcs=[], package_name="nx_parallel"):
 
 
 def test_get_functions_with_get_chunks():
-    expected = {attr.rsplit(".", 1)[-1] for attr in ALGORITHMS}
+    expected = {algo.rsplit(".", 1)[-1] for algo in ALGORITHMS}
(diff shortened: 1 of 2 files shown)


nx_parallel/tests/test_get_chunks.py
```diff
@@ -30,30 +31,7 @@ def get_functions_with_get_chunks(ignore_funcs=[], package_name="nx_parallel"):
 
 
 def test_get_functions_with_get_chunks():
-    # TODO: Instead of `expected` use ALGORTHMS from interface.py
-    # take care of functions like `connectivity.all_pairs_node_connectivity`
```
Schefflera-Arboricola, line 34: By "functions like `connectivity.all_pairs_node_connectivity`," I meant those with a `name` parameter defined in the `_dispatchable` decorator. For example, `approximate_all_pairs_node_connectivity` is not the actual name of the function in the networkx but in the `name` parameter of the decorator-- because the name `all_pairs_node_connectivity` is already used by `connectivity.all_pairs_node_connectivity`. The current logic-- splitting on `.` -- works with existing functions, but i'm not sure if there's a consistent rule for setting the `name` parameter in networkx. It may be worth updating the code if there is a rule or adding a comment to clarify this for future reference/maintenance.

akshitasure12: Thank you! That makes sense — I’ll close this and keep it in mind to revisit when needed :)

nx_parallel/tests/test_get_chunks.py
```diff
@@ -30,31 +31,7 @@ def get_functions_with_get_chunks(ignore_funcs=[], package_name="nx_parallel"):
...
-        "is_reachable",
-        "johnson",
-        "local_efficiency",
-        "node_redundancy",
-        "number_of_isolates",
-        "square_clustering",
-        "tournament_is_strongly_connected",
-    }
-    assert set(get_functions_with_get_chunks()) == expected
+    assert set(get_functions_with_get_chunks()) == set(ALGORITHMS)
```
Schefflera-Arboricola, line 34:
```suggestion
    assert get_functions_with_get_chunks() == ALGORITHMS
```

PR #130, opened 2025-07-29
parallel implementation of `clustering` and `average_clustering`

akshitasure12:
This PR implements the [`clustering`](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cluster.clustering.html) and [`average_clustering`](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cluster.average_clustering.html)

Note: I ran the timing script for 1–1.5 hours, but it failed to produce results for 1,600-node graphs-- the current heatmap includes speedups only up to 800 nodes. Please let me know if the implementation is missing something, thank you!


akshitasure12:
I explored three approaches for implementing `average_clustering`:
1. Parallelise `average_clustering` while using NetworkX’s sequential `clustering`.
2. Use NetworkX’s `average_clustering` but plug in the parallel `clustering` implementation.
3. Parallelise `average_clustering` and also use parallel `clustering` (nested parallelism).

Running the timing script showed that (1) and (2) achieved similar speedups. (ref [heatmaps]()).

I discarded (3) due to the risks of nested parallelism and oversubscription of CPU cores (ref. [nested parallelism](https://stackoverflow.com/questions/4317551/openmp-what-is-the-benefit-of-nesting-parallelizations)) While (3) produced similar speedups to (1) and (2) in my tests, this was more of a precautionary decision to avoid performance degradation for graphs with much larger node counts. 

Between the remaining two, I chose (1) as it best aligns with nx-parallel’s design philosophy of applying parallelism at the top-level algorithm rather than within nested calls. lmkwyt :)

nx_parallel/algorithms/cluster.py
```diff
@@ -70,3 +80,123 @@ def _compute_clustering_chunk(node_iter_chunk):
...
+        G = G.graph_object
+
+    n_jobs = nxp.get_n_jobs()
+
+    if nodes is None:
+        nodes_to_chunk = list(G)
+    elif isinstance(nodes, str) or not isinstance(nodes, Iterable):
+        nodes_to_chunk = [nodes]
+    else:
+        nodes_to_chunk = list(nodes)
```
dschult, line 139: Can this if structure be accomplished using `nodes_to_chunk = list(G.nbunch_iter(nodes))`?
If not, can  the middle check use something like: `elif hashable(nodes):`? What is the middle check supposed to handle?

dschult (commented):
This looks good! Just one question about the handling of the `nodes` parameter,

nx_parallel/algorithms/cluster.py
akshitasure12: My original goal was simply to handle both single nodes and iterables, but this approach makes it much cleaner. I’ll switch to `list(G.nbunch_iter(nodes))` since it already handles all those cases directly. Thanks @dschult for pointing it out!

dschult (approved):
Looks good!

PR #141, opened 2025-08-22
Add `should_run` policy for when `nodes` is `None`

### LINKED ISSUE #137: set `should_run=False` unless `nodes` is `None`
Actually, I think this might be a case when we could implement `should_run` to return False unless `nodes is None`. But that can be for a later PR.

_Originally posted by @dschult in https://github.com/networkx/nx-parallel/pull/106#pullrequestreview-3125965245_

akshitasure12:
Fixes #137

### DIFF
--- a/nx_parallel/tests/test_should_run.py
+++ b/nx_parallel/tests/test_should_run.py
@@ -52,6 +52,19 @@ def test_should_run_if_large():
     assert dummy_if_large.should_run(largeG)
 
 
+def test_should_run_if_nodes_none():
+    @nxp._configure_if_nx_active(should_run=nxp.should_run_if_nodes_none)
+    def dummy_nodes_none(G, nodes=None):
+        pass
+
+    G = nx.fast_gnp_random_graph(20, 0.6, seed=42)
+    assert (
+        dummy_nodes_none.should_run(G, nodes=[1, 3])
+        == "`nodes` should be None for parallel execution"
+    )
+    assert dummy_nodes_none.should_run(G)
+
+
 @pytest.mark.parametrize("func_name", get_functions_with_should_run())
 def test_should_run(func_name):
     tournament_funcs = [
--- a/nx_parallel/utils/should_run_policies.py
+++ b/nx_parallel/utils/should_run_policies.py
@@ -5,6 +5,7 @@ __all__ = [
     "default_should_run",
     "should_skip_parallel",
     "should_run_if_large",
+    "should_run_if_nodes_none",
 ]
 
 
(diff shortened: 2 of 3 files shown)


Schefflera-Arboricola (commented):
If the parallel implementation can only run with `nodes=None` then I think `can_run` would be more appropriate here-- instead of should_run. Also, please always remember to document such details. Thanks!

nx_parallel/tests/test_should_run.py
```diff
@@ -52,6 +52,19 @@ def dummy_if_large(G):
...
+        pass
+
+    G = nx.fast_gnp_random_graph(20, 0.6, seed=42)
+    assert (
+        dummy_nodes_none.should_run(G, nodes=[1, 3])
+        == "`nodes` should be None for parallel execution"
+    )
+    assert dummy_nodes_none.should_run(G)
+
+
```
dschult, line 67:
```suggestion
 
```

nx_parallel/utils/should_run_policies.py
```diff
@@ -27,6 +28,12 @@ def default_should_run(*_):
     return True
 
 
+def should_run_if_nodes_none(G, nodes=None, *_):
+    if nodes is None:
+        return True
+    return "`nodes` should be None for parallel execution"
```
dschult, line 34:
```suggestion
    return "Parallel execution only supported when `nodes` is None"
```

nx_parallel/tests/test_should_run.py
```diff
@@ -52,6 +52,19 @@ def dummy_if_large(G):
...
 
+def test_should_run_if_nodes_none():
+    @nxp._configure_if_nx_active(should_run=nxp.should_run_if_nodes_none)
+    def dummy_nodes_none(G, nodes=None):
+        pass
+
+    G = nx.fast_gnp_random_graph(20, 0.6, seed=42)
+    assert (
+        dummy_nodes_none.should_run(G, nodes=[1, 3])
+        == "`nodes` should be None for parallel execution"
```
dschult, line 63:
```suggestion
        == "Parallel execution only supported when `nodes` is None"
```

dschult:
The test failure shows the difference between "should_run" and "can_run". :)
We should leave the code in place, because while "should_run" says we should not run it, we still can (by specifying the backend parameters for example). If we want to remove that code we should make a `can_run` function that returns False in this case.   Personally, I think we can leave the code in place and provide the should_run. But I can imagine others feeling differently.  And we might change our minds on the approach in the future (still developing these kinds of traditions).

nx_parallel/tests/test_should_run.py
Schefflera-Arboricola: For a consistent and organized test structure, it would be good to split the tests in this file-- tests that test `should_run` used in nx-parallel can remain in this file, while tests that test the functionality implemented in `should_run_policies.py` (i.e. tests with dummy functions, data, etc.) can live in `utils/tests/test_should_run_policies.py`. (like `get_chunks`)

PR #145, opened 2025-12-13
Added functionality of parallel maximal independent set

### LINKED ISSUE #144: Add parallel implementation of Combinatorial Algorithms
NetworkX currently implements many core combinatorial graph primitives in a **strictly sequential** manner. A clear example is `maximal_independent_set`, whose current implementation is a randomized **serial greedy** algorithm. This approach is inherently non-parallelizable and does not reflect the state of the art in parallel/distributed graph processing.

Since `nx-parallel` aims to provide scalable, concurrency-friendly variants of NetworkX algorithms, it makes sense to introduce **parallel implementations** of classical combinatorial routines. A foundational starting point is a proper parallel MIS algorithm.

## **The existing NetworkX MIS is sequential**

The current implementation -> https://networkx.org/documentation/stable/_modules/networkx/algorithms/mis.html#maximal_independent_set:

```python
available = V \ (MIS ∪ N(MIS))
while available:
    v = random choice from available
    add v to MIS
    remove v and N(v) from available
```

Key issues:

1. **Exactly one vertex is added per iteration.**
   The algorithm performs a strictly serial “pick-one-and-delete-its-neighbors” loop. There is no mechanism for selecting a batch of independent vertices simultaneously.

2. **Algorithmic structure = greedy MIS under a random vertex ordering.**
   Conceptually, this corresponds to traversing one random permutation π of V and inserting a vertex whenever none of its previously visited neighbors are in MIS. This inherently imposes a total order.

3. **Neighborhood checks are performed on the updated graph state.**
   After each vertex is added, all its neighbors are removed before the next selection. This makes the next step depend on the results of previous steps, preventing any parallel independence checks.

4. **No use of random priorities.**
   Parallel MIS algorithms (Luby, Blelloch-Halperin, Bader-Cong, ECL-MIS, etc.) rely on *random priority assignments* to allow many vertices to compare themselves with neighbors **simultaneously**.
   The existing implementation does not expose any priority vector that can be evaluated in parallel.

As a result, the current MIS is inherently **sequential both in logic and data dependencies**.

## **Parallelize MIS (Luby-style approach)**

A standard parallel MIS algorithm assigns a random priority to **every vertex once per round**, then selects **all locally maximal vertices at the same time**:

```
while graph not empty:
    generate random priority for each vertex
    S = { u | priority(u) > priority(v) for all v in N(u) }
    add S to MIS
    remove S ∪ N(S) from graph
```

Why this is parallel:

* Priority comparisons for all vertices are independent operations.
* Set S can be selected in one parallel pass.
* All vertices in S are guaranteed independent.
* Removing S ∪ N(S) can also be done in parallel.

This matches the PRAM formulation and is widely used in CPU and GPU environments.

PS I am new to nx-parallel, however, I have implemented parallel variants of various combinatorial algorithms such as MIS, Graph Coloring, Matching etc earlier. I think adding this will be a valuable addition to nx-parallel.

Let me know your thoughts.

pnijhara:
## What does this PR change?
This PR implements the parallel maximal independent set via Luby's random permutation mechanism (solving issue #144)
- Add parallel implementation of maximal_independent_set algorithm
- Implements Luby-style randomized parallel algorithm for speedup on large graphs (50K+ nodes)
- Works best on large graphs (n>50,000)

## What is your approach, and why did you choose it?
A standard parallel MIS algorithm assigns a random priority to **every vertex once per round**, then selects **all locally maximal vertices at the same time**:

```
while graph not empty:
    generate random priority for each vertex
    S = { u | priority(u) > priority(v) for all v in N(u) }
    add S to MIS
    remove S ∪ N(S) from graph
```

Why this is parallel:

* Priority comparisons for all vertices are independent operations.
* Set S can be selected in one parallel pass.
* All vertices in S are guaranteed independent.
* Removing S ∪ N(S) can also be done in parallel.

This matches the PRAM formulation and is widely used in CPU and GPU environments.

## Brief summary of changes (1–2 lines)
- Added a file algorithms/mis.py
- Added tests for algorithms/mis.py in algorithms/tests/test_mis.py
- Found in the local benchmark tests that for node count less than 50K, sequential code is working best; however, for larger graphs such as n = 100K and 1M, parallel code is working best. This may be because of joblib's parallelization overhead. I tested this in C++ with OpenMP. The same issue persists there, but with a lesser node count.


dschult (commented):
Thanks for this!

Some formatting suggestions before diving into the code/tests:
- we typically don't put error messages on the "assert" command. The pytest report of the assert failure is sufficient.
- we also typically dont include a comment line at the beginning of the test when the test function's name is sufficient. If the test involves soemthing special or unexpected, then a couple line doc_string (or comment) is used. We aren't building docs for the tests.
- instead of building a sequential version of the code inside nx-parallel, does it make sense to just "fall back" to the NetworkX version?

pnijhara:
I tried as suggested, I am not very sure whether I have done this on correct side. Now, instead of duplicating NetworkX's sequential MIS code, we now fall back to their implementation for small graphs (< 50,000 nodes). 

What I did:

1. Dual-mode ``should_run_policies.py:16-41` policy

Modified to work in two modes:
- Direct function: `should_run=nxp.should_run_if_large` uses default 200 node threshold
- Factory function: `should_run=nxp.should_run_if_large(50000)` returns wrapper with custom threshold

Also updated all policy signatures to accept `**__` for keyword arguments passed by the backend dispatcher.

2. Unwrapping NetworkX's dispatcher
Used `inspect.unwrap()` to access the actual NetworkX implementation, bypassing multiple decorator layers. Direct import caused infinite recursion since NetworkX's `maximal_independent_set` is itself a dispatched function. This was hard to fix. I had to use LLMs, sorry :( 

3. Explicit fallback check

Added manual `should_run` check because NetworkX bypasses `should_run` when backend is explicitly specified (e.g., `backend='parallel'`). Also converted seed to `Random` object before fallback since the unwrapped NetworkX function expects `Random`, not `int`.

dschult (commented):
I haven't worked through the actual code yet, but looked at the tests and the docs. The tests seem like maybe some of them are already covered when we run the networkx tests using the environment variable set for nx-parallel.  In that case, each networkx test is run with the nx-parallel code being called when it is supported. And we run that in  the nx-parallel CI tests.

- can you see whether any of the nx-parallel tests are already tested by the networkx tests -- and remove the duplicates?
- I have a couple other comments/questions below.

I hope to look at the code itself soon.
:)

nx_parallel/utils/should_run_policies.py
```diff
@@ -9,20 +9,39 @@
...
+
+        if len(G) < threshold:
+            return "Graph too small for parallel execution"
+        return True
+
+    # Otherwise, it's being used as a factory, return a wrapper
+    threshold = nodes_threshold
+    def wrapper(G, *_, **__):
+        if hasattr(G, "graph_object"):
+            G = G.graph_object
```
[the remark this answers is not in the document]
pnijhara: Precisely yes.

nx_parallel/utils/should_run_policies.py
```diff
@@ -9,20 +9,39 @@
...
+def should_skip_parallel(*_, **__):
     return "Fast algorithm; skip parallel execution"
 
 
-def should_run_if_large(G, *_):
-    if hasattr(G, "graph_object"):
-        G = G.graph_object
+def should_run_if_large(nodes_threshold=200, *_, **__):
+    # If nodes_threshold is a graph-like object, it's being used as a direct should_run
+    # function instead of a factory. Use default threshold.
```
dschult: I would prefer the function signature here be: `should_run_if_large(G, nodes_threshold=200, *_, **__)`
So can you briefly explain where this might be used as a factory? The previous version doesn't seem to deal with that case. Of course, it didn't deal with keyword args either...   ;}

nx_parallel/algorithms/mis.py
```diff
@@ -0,0 +1,193 @@
...
+
+    # Parallel strategy: Run complete MIS algorithm on node chunks independently
+    # Then merge results by resolving conflicts
+    all_nodes = list(G.nodes())
+
+    # Remove required nodes and their neighbors from consideration
+    if nodes_set:
+        available = set(all_nodes) - nodes_set
+        for node in nodes_set:
+            available.discard(node)
```
dschult: Dont these nodes already get removed two lines up?

nx_parallel/algorithms/mis.py
```diff
@@ -0,0 +1,193 @@
...
+        for i, chunk in enumerate(chunks)
+    )
+
+    # Merge results: resolve conflicts between chunks
+    indep_set = list(nodes_set) if nodes_set else []
+    excluded = set(nodes_set)
+
+    if nodes_set:
+        for node in nodes_set:
+            excluded.update(adj_dict[node])
```
dschult: Didn't we do this already with `available`?

nx_parallel/algorithms/mis.py
```diff
@@ -0,0 +1,193 @@
...
+    provides speedup on large graphs (>= 50000 nodes). For smaller graphs,
+    the NetworkX sequential version is used automatically.
+
+    """
+    if hasattr(G, "graph_object"):
+        G = G.graph_object
+
+    # Validate directed graph
+    if G.is_directed():
+        raise nx.NetworkXNotImplemented("Not implemented for directed graphs.")
```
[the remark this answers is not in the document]
pnijhara: NetworkX uses @not_implemented_for("directed") decorator (line 12 in networkx/networkx/algorithms/mis.py). This decorator runs before backend dispatch because it's the outermost decorator.

For our nx-parallel version I used a manual check at lines 84-86 that runs inside the function body, after the backend has been dispatched.

nx_parallel/algorithms/mis.py
pnijhara: Removed the redundant available.discard(node). Since `available = set(all_nodes) - nodes_set` already excludes nodes in nodes_set, the discard is unnecessary. The loop now only removes neighbors of the required nodes.

nx_parallel/utils/should_run_policies.py
pnijhara: I have slightly updated the code now:

```python
def should_run_if_large(G=None, nodes_threshold=200, *_, **__):
    # Detect if first arg is a graph (has both __len__ and nodes attributes)
    is_graph = G is not None and hasattr(G, '__len__') and hasattr(G, 'nodes')
    
    if is_graph:
        # Direct usage: called with a graph as first argument
        # Example: should_run_if_large(G) or func.should_run(G)
            G = G.graph_object

        if len(G) < nodes_threshold:
            return "Graph too small for parallel execution"
        return True
    
    # Factory usage: called with threshold (positional or keyword) but no graph
    # Examples: should_run_if_large(50000) or should_run_if_large(nodes_threshold=50000)
    # Use G if it's a number (threshold passed positionally), otherwise use nodes_threshold
    threshold = G if G is not None and isinstance(G, (int, float)) else nodes_threshold
    
    def wrapper(G, *_, **__):
            G = G.graph_object

        if len(G) < threshold:
            return "Graph too small for parallel execution"
        return True

    return wrapper
```

The function is used as a factory when you want to create a should_run function with a custom threshold:
`@nxp._configure_if_nx_active(should_run=nxp.should_run_if_large(50000))`

Here, we called the `should_run_if_large(50000)` with a threshold, returning a wrapper function that uses that threshold. The decorator stores this wrapper as the should_run method.

pnijhara:
@dschult thanks for the detailed review. I made the suggested changes and also tried to defend myself in some cases :p
Please review it again.

nx_parallel/algorithms/mis.py
```diff
@@ -0,0 +1,189 @@
+from typing import Any
+from joblib import Parallel, delayed
+import nx_parallel as nxp
+import networkx as nx
+
+__all__ = ["maximal_independent_set"]
+
+# Import the actual NetworkX implementation
+from networkx.algorithms.mis import maximal_independent_set as _nx_mis_dispatcher
+_nx_mis = _nx_mis_dispatcher.orig_func
```
dschult: I think this can be simpler -- not even an import. And since we only use it once, maybe we can just use the expression directly without the abbreviation.
```python
# abbreviation for the actual NetworkX implementation
_nx_mis = nx.maximal_independent_set.orig_func
```

nx_parallel/algorithms/mis.py
```diff
@@ -0,0 +1,189 @@
+from typing import Any
```
dschult: This isnt used

pnijhara:
@dschult 
Can you please review this PR again?

dschult (commented):
It looks like the style test is not passing. 
I have two more import related comments.

I think the tests should focus on aspects of the nx-parallel implementation. All the NetworkX tests are run on this function -- so don't duplicate those. You can add tests to the networkx tests if you like. :)  but not in this PR. :)

And I think you still haven't responded to the comment about code that is inside a "should_run" check.  The should_run gets called by the networkx dispatcher so we should never get to the check here in the backend.

pnijhara:
@dschult tests are failing because the parallel algorithm produces a valid MIS [4, 3, 2, 0, 1] but in a different order than the sequential [1, 0, 3, 2, 4]. This is expected behavior for parallel execution. The question is: should the parallel backend match sequential ordering exactly, or is correctness (independent + maximal) sufficient?

We can either accept the test failure as expected (parallel produces correct but different results) or make small graphs fall back to sequential, even when `NETWORKX_TEST_BACKEND` is set.

dschult:
We should be testing correctness. So we probably need a small change to the NetworkX test. I think that test should be updated in NetworkX to accept any ordering -- probably by comparing the set of the output to an expected value set.

There may be a way for backends to avoid NetworkX tests that give different results, but I don't know it. @eriknw, is there a way to e.g. mark a networkx test as xfail when running it on a backend?  

In this case though, I think a small PR to change line 15 of `test_mis.py` to:
```python
    assert set(nx.maximal_independent_set(G, seed=1)) == {1, 0, 3, 2, 4}
```

Also, [there is even talk in NetworkX](https://github.com/networkx/networkx/issues/8554) about whether mis should return a list or a set...  But It'd be good to have this changed.

eriknw:
Thanks for the ping!

> I think that test should be updated in NetworkX to accept any ordering

Yeah, for easy changes, I favor tweaking tests in NetworkX. Moreover, if a backend add _new_ tests to test for gaps in NX tests, I prefer to upstream those to so NetworkX and all backends benefit. One challenge with this approach though is that updated tests are part of the _next_ NetworkX release, and these tests may still fail _now_.

> There may be a way for backends to avoid NetworkX tests that give different results, but I don't know it. @eriknw, is there a way to e.g. mark a networkx test as xfail when running it on a backend?

`on_start_tests` can do this. It is minimally documented: https://networkx.org/documentation/stable/reference/backends.html#creating-a-custom-backend

That doc should also probably refer to https://docs.pytest.org/en/stable/reference/reference.html#std-hook-pytest_collection_modifyitems

You can see how `nx-cugraph` uses `on_start_tests` to skip tests here: https://github.com/rapidsai/nx-cugraph/blob/10426797772a3e99bd45cf3cfa96c7ae7184431c/nx_cugraph/interface.py#L33

pnijhara:
@dschult let me know what decision we are taking for this? Ideally, we should update the NetworkX tests. However, those will take time to get into the final release. Three things we can do now:
1. Keep the failing tests (not an ideal solution)
2. I should send a PR on NetworkX to update the tests
3. As suggested in the previous comment, I can edit the `interface.py` file for `on_start_tests` in the `BackendInterface` class. Something like this:

```python
@staticmethod
def on_start_tests(items):
    """Modify pytest items after tests have been collected.
    
    This is called during pytest_collection_modifyitems phase.
    """
    try:
        import pytest
    except ModuleNotFoundError:
        return
    
    def key(testpath):
        """Parse testpath into (test_name, keywords) tuple."""
        filename, path = testpath.split(":")
        *names, testname = path.split(".")
        if names:
            [classname] = names
            return (testname, frozenset({classname, filename}))
        return (testname, frozenset({filename}))
    
    # Define tests that should be marked as xfail
    xfail = {
        key("test_mis.py:TestMaximalIndependentSet.test_random_order"):
            "Parallel MIS produces different valid results than sequential",
        key("test_mis.py:TestMaximalIndependentSet.test_seed_deterministic"):
            "Parallel MIS with chunks produces different order than sequential",
    }
    
    # Mark tests
    for item in items:
        kset = set(item.keywords)
        for test_name, reason in xfail.items():
            test_name_val, keywords = test_name
            if item.name == test_name_val and keywords.issubset(kset):
                item.add_marker(pytest.mark.xfail(reason=reason))
```

dschult:
Nice!  Let's do numbers 2 and 3.
Or, maybe let's focus on 3 here, and hopefully the issue/pr in NetworkX to refactor the mis return value from a list to a set will fix the other issues.  

So, work on 3. And if by the time we've finished that here, we still need to do 2, we can do that then.
Thanks!

pnijhara:
@dschult, I just implemented and tested 3. Looks like it is working fine. If this is good to go, I can open a new issue at NetworkX for solving 2.


PS: Please ignore the reviews by Copilot. I didn't know it can review the PRs. I got it today with my education pack. Seems like a too much to me.

pnijhara:
@dschult, can you review and merge this? I think the tests are passing. We can revert these tests once we get the new release of NetworkX.

nx_parallel/algorithms/mis.py
```diff
@@ -0,0 +1,189 @@
...
+    rng.shuffle(available)
+
+    # Split into chunks
+    if get_chunks == "chunks":
+        chunks = list(nxp.chunks(available, n_jobs))
+    else:
+        chunks = list(get_chunks(available))
+
+    # Precompute adjacency
+    adj_dict = {node: set(G.neighbors(node)) for node in G.nodes()}
```
dschult, line 137: Why can't we use `G.neighbors` inside the chunk processing routine?
We are basically copying the entire graph here. 

Secondly, why do we need to make this a set? It can be any type of iterable to be useful in the chunk processing func and also in the `update` lines below. What do we need?

nx_parallel/algorithms/mis.py
```diff
@@ -0,0 +1,189 @@
...
+    >>> import nx_parallel as nxp
+    >>> G = nx.path_graph(5)
+    >>> nxp.maximal_independent_set(G)  # doctest: +SKIP
+    [4, 0, 2]
+    >>> nxp.maximal_independent_set(G, [1])  # doctest: +SKIP
+    [1, 3]
+
+    Notes
+    -----
+    This algorithm does not solve the maximum independent set problem.
```
dschult, line 71: If we include this sentence in the notes section we need to explain what the difference is. And it probably isn't related to the chunk based parallel algorithm sentence immediately following. So that should be in a separate paragraph.

dschult (commented):
Do you have any timing info on this function?
How is it doing? How did we choose 50000 nodes for should run?

dschult (approved):
I think this is ready. Are there other things ou'd like to do before it is merged?
Anything we should do as a followup?

I approve this PR

pnijhara:
> Anything we should do as a follow-up?

The only thing that I didn't answer till now is how that 5000 as a number came up. For that, in January, I had access to a server-grade machine with 128 cores, there on 20 chunks, for an ER or Power law graph of vertices more than 50K with decent sparsity, I was getting speedups from 1.5x to 4x.

Though I cannot verify these numbers again, as I don't have any access to that machine. Apart from that, everything seems fine to me. Parallelism is there. On small graphs, our parallel implementation won't win because chunks has there own overheads.

Thank you, @dschult, for reviewing this PR. It is indeed one of the longest running PR I have worked on.

PR #154, opened 2026-01-02
FIX: Respect Joblib context in config sync (Issue #76)

### LINKED ISSUE #76: Synchronizing NetworkX and Joblib configurations in nx-parallel
## Background:

PR https://github.com/networkx/nx-parallel/pull/75 adds implementation to make nx-parallel compatible with the networkx's configuration system, offered for networkx backends, and documents how to configure nx-parallel using joblib’s `parallel_config`. 

PR 68 was an attempt to create a unified configuration system in `nx-parallel` by wrapping the `joblib.Parallel()`(inside nx-parallel algorithms) within a `with joblib.parallel_config(configs)` context, here `configs` are extracted from `nx.config.backends.parallel`. This approach made NetworkX’s config closely mirror joblib’s, giving the appearance of synchronization between the two systems. However, in the last meeting, Dan mentioned that this approach complicates things.

## Requirement:

We need a robust layer in `nx-parallel` which updates NetworkX’s config as soon as any of joblib’s config parameters are updated, and vice versa. The goal is to ensure that when a user updates a configuration in either of the systems, the other system is instantaneously updated by this layer/interface in nx-parallel to maintain consistency.

## Challenges:

1. **Order of Updates:**
   - As an `nx-parallel` developer, how can I determine which config system (NetworkX or joblib) the user has updated and in what order?
   - For example, consider the following:
     ```python
     joblib.parallel_config(n_jobs=5)
     nx.config.backends.parallel.n_jobs = 8
  
     nx.config.backends.parallel.n_jobs = 8
     joblib.parallel_config(n_jobs=5)
     ```
     - It’s simple to extract config stored in networkx and in joblib’s `parallel_config`. So, in both cases, we will extract `n_jobs = 8` from NetworkX and `n_jobs = 5` from joblib. However, the expected behavior is different: in the first case, `n_jobs` should be 8 at the end, and in the second case, it should be 5. How can I detect the order of updates to apply the correct configuration?

2. **Keeping Context Managers in Sync:**
   - Is there a way in Python to determine the context manager we are currently in and the order in which multiple context managers are applied?
   - Consider the following:
     ```python
     nx.config.backends.parallel.n_jobs = 6 
     joblib.parallel_config(verbose=50) 
     G = nx.complete_graph(5) 
     nx.square_clustering(G, backend="parallel")  # n_jobs = 6, verbose = 50 
     
     with nx.config.backends.parallel(n_jobs=8): 
         nx.square_clustering(G, backend="parallel")  # n_jobs = 8, verbose = 50 
         with joblib.parallel_config(verbose=10): 
             nx.square_clustering(G, backend="parallel")  # n_jobs = 8, verbose = 10
     ```
   - How can we ensure that the configurations within nested context managers are correctly synchronized and applied in the expected order?

3. **Updating Configurations Without Modifying External Libraries?:**
   - Is it even possible to update both NetworkX’s and joblib’s configurations to keep them in sync without requiring any code changes in either of the libraries?

Thank you :)

ANJAN672:
Fixes #76 (partial)

## Problem
When using `joblib.parallel_config(n_jobs=X)`, nx-parallel was ignoring the value and always using `nx.config.backends.parallel.n_jobs`.

## Solution
- Modified [decorators.py](cci:7://file:///d:/nx-parallel/nx_parallel/utils/decorators.py:0:0-0:0) to peek at Joblib's active context before overriding
- Modified [chunk.py](cci:7://file:///d:/nx-parallel/nx_parallel/utils/chunk.py:0:0-0:0) to prioritize Joblib's active value over NX config

## Known Limitation
When NX context is nested inside Joblib context, Joblib still wins. Full sync requires upstream changes to both libraries.

## Tests
All 89 existing tests pass.

### DIFF
--- a/nx_parallel/utils/decorators.py
+++ b/nx_parallel/utils/decorators.py
@@ -3,6 +3,7 @@ from dataclasses import asdict
 from functools import wraps
 import networkx as nx
 from joblib import parallel_config
+from joblib.parallel import get_active_backend
 from nx_parallel.utils.should_run_policies import default_should_run
 
 
@@ -21,11 +22,18 @@ def _configure_if_nx_active(should_run=None):
                 nx.config.backends.parallel.active
                 or "PYTEST_CURRENT_TEST" in os.environ
             ):
-                # Activate nx config system in nx_parallel with:
-                # `nx.config.backends.parallel.active = True`
+                # Peeking at joblib's current state to see if there's an outer context
+                _, current_n_jobs = get_active_backend()
+
+                # Get NetworkX config
                 config_dict = asdict(nx.config.backends.parallel)
                 config_dict.update(config_dict.pop("backend_params"))
                 config_dict.pop("active", None)
+
+                # SYNC: If user has an outer Joblib context for n_jobs, respect it!
+                if current_n_jobs is not None:
+                    config_dict["n_jobs"] = current_n_jobs
+
                 with parallel_config(**config_dict):
                     return func(*args, **kwargs)
             return func(*args, **kwargs)
--- a/nx_parallel/tests/test_should_run.py
+++ b/nx_parallel/tests/test_should_run.py
@@ -3,7 +3,6 @@ from nx_parallel.interface import ALGORITHMS
 import networkx as nx
 import inspect
 import pytest
-import os
 
 
 def get_functions_with_should_run():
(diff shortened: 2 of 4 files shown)


nx_parallel/utils/decorators.py
```diff
@@ -21,11 +22,18 @@ def wrapper(*args, **kwargs):
...
+                # Peeking at joblib's current state to see if there's an outer context
+                _, current_n_jobs = get_active_backend()
+
+                # Get NetworkX config
                 config_dict = asdict(nx.config.backends.parallel)
                 config_dict.update(config_dict.pop("backend_params"))
                 config_dict.pop("active", None)
+
+                # SYNC: If user has an outer Joblib context for n_jobs, respect it!
+                if current_n_jobs is not None:
```
Schefflera-Arboricola, line 34: what if other parameters (other than n_jobs) were changed? -- we need to sync the entire context manager classes -- along with the default value mismatches (like for n_jobs and backend_args dict)

Schefflera-Arboricola (commented):
Thanks for the PR @ANJAN672 !

> When using joblib.parallel_config(n_jobs=X), nx-parallel was ignoring the value and always using nx.config.backends.parallel.n_jobs

you can change that by setting `active=False` in nx.config , read - https://github.com/networkx/nx-parallel/blob/main/Config.md

Also read - https://github.com/networkx/nx-parallel/issues/76#issuecomment-3705551886 -- And please consider closing this PR if you agree with it. Thx!

nx_parallel/utils/decorators.py
ANJAN672: Thanks for the detailed explanation @Schefflera-Arboricola! 

After reading your comment on the issue, I understand the design rationale now - keeping the two systems separate with the [active](cci:1://file:///d:/nx-parallel/nx_parallel/utils/decorators.py:12:0-46:20) flag is cleaner than implicit syncing.

I learned a lot about the nx-parallel architecture while working on this. Closing this PR as per your suggestion.

Looking forward to contributing more to nx-parallel!

PR #156, opened 2026-01-11
Add latapy_clustering parallel algorithm in bipartite

gulshan-123:
## What does this PR change?
- Implement the latapy clustering of the bipartite graph as parallel.
- Also noticed a minor bug in timing script, where while ensuring the neighbour size >= 2, the resulting graph may not be bipartite.

## Brief summary of changes (1–2 lines)
For the second issue, I've used the fact that `nx.bipartite.random_graph` partitions the nodes into two parts as follows:
- the first part has node indices 0..n-1
- the second part has node indices n..n+m-1

## Additional comments/resources or notes for reviewers
https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.bipartite.generators.random_graph.html

EDIT: 
- [x] I have to edit process_chunk for tests to complete.

cc @Schefflera-Arboricola 

---
(In future, I wish to complete #72 : parallel floyd warshall specifically)

PR #158, opened 2026-03-04
Timing script for bipartite actually producing bipartite graph

gulshan-123:
As suggested by @akshitasure12 in #156  , the change to timing script is in this PR.

Earlier, the bipartite graph part was not actually producing bipartite graph, this was noticed when the graphs were returning error in latapy_clustering tests.

I have used the fact that nx.bipartite.random_graph partitions the nodes into two parts as follows:

    the first part has node indices 0..n-1
    the second part has node indices n..n+m-1


gulshan-123:
Hi @akshitasure12, Is this PR and #156, good to go, or need any changes?

akshitasure12:
looks good to me!

dschult (approved):

PR #160, opened 2026-08-15
Added a comprehensive tutorial notebook for all supported parallel backends (with free-threaded python) and overhauled the readme

Schefflera-Arboricola:


Schefflera-Arboricola:
@dschult -- just a gentle ping. Thx!

dschult (approved):
Sorry for the delay on reviewing this.
It looks good!  I think the changes to README are uniformly improvements.  :)

The tutorial notebook looks good for content. I haven't checked that everything runs as expected/documented. But that breaking would be strange. And I haven't tried installing what I would need to get this running live in a jupyter sessions. But the static version of the cells looks good. 

I think this shifts the focus of the readme a little from running specific algorithms to describing how the package works with the joblib options for different parallel systems.  I think this is a reasonable shift matching the packages maturity -- going beyond a few functions to many functions and now via different parallel systems.

Thansk @Schefflera-Arboricola for putting this together!!

